PR #429, opened 2014-12-01
Subdir

mattmb:
Allows us to host the app on a subdirectory.

This will be /employment-tribunals. Or whatever's set as RAILS_RELATIVE_URL_ROOT.

Unfortunately we have a bunch of static links in our locale file. Is it rendered anywhere i.e. can we use the environment variable somehow?

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,47 +1,49 @@
 Rails.application.routes.draw do
+  scope ENV['RAILS_RELATIVE_URL_ROOT'] || '/' do
 
-  resource :guide, only: :show
+    resource :guide, only: :show
 
-  resource :terms, only: :show
+    resource :terms, only: :show
 
-  resource :user_session, only: %i<create update>, path: :application do
-    member do
-      %w<reminder returning refresh-session session-expired>.each do |page|
-        get page
-      end
-    end
-  end
-
-  scope :apply do
-    resource :claim_review, only: %i<show update>, path: :review
-
-    resource :claim_confirmation, only: :show, path: :confirmation do
-      get 'generated_claim', on: :member
-    end
-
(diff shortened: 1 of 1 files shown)


stevegraham:
I spoke to @tomwm before xmas and our app will be launched in parallel to Jadu's remaining in service, so ours can't be mounted at `/employment-tribunals` as this is taken by Jadu's.

I will close this after speaking to @helenmott, I remember there being another reason why we might need a prefix so I want to check first. If we need to maintain access to legacy pages we can proxy them in nginx.

helenmott:
The other reason was there are a number of forms that sit within the
Employment Tribunal service (ours, the ET3 and the EAT), and each of these
would need their own URL.

On 5 January 2015 at 10:15, Stevie Graham <notifications@github.com> wrote:

> I spoke to @tomwm <https://github.com/tomwm> before xmas and our app will
> be launched in parallel to Jadu's remaining in service, so ours can't be
> mounted at /employment-tribunals as this is taken by Jadu's.
>
> I will close this after speaking to @helenmott
> <https://github.com/helenmott>, I remember there being another reason why
> we might need a prefix so I want to check first. If we need to maintain
> access to legacy pages we can proxy them in nginx.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ministryofjustice/atet/pull/429#issuecomment-68689341>
> .
>



-- 
Helen Mott
Head of Clan for Delivery
MoJ Digital Services
07712 667792

stevegraham:
let's discuss after standup.

mattmb:
@stevegraham I think there's also a connection with the links on gov.uk?

Also if there's going to be some period when we do coexist with Jadu then we'll still need this. Will be in after lunch if you want to discus.

PR #494, opened 2014-12-15
Use correct filename when sending to Jadu

mattmb:
@dan-palmer think this was the fix so that the correct filename gets into the XML. Not sure if you want to fix the problem with the underscores too or wait to hear back from Jadu.


mattmb:
@dan-palmer are you on this? It's the next blocker to an e2e test.

dan-palmer:
@mattmb yeah, working on a fix that is consistent across the app

PR #506, opened 2014-12-16
Update carrierwave version

mattmb:
This version includes this fix:
https://github.com/carrierwaveuploader/carrierwave/pull/1517

There's no recent release so I've just pinned to the latest commit on master. Worryingly the carrierwave build is failing but the fix was only merged a few days ago...

@stevegraham @dan-palmer


mattmb:
@dan-palmer can you take a look at that test failure? need to get this merged so we can do our e2e test.

PR #523, opened 2014-12-30
Front end/ga virtual page views

cleargif:
This is done. Will need to review it with Trent and possibly make tweaks.

Did a rebase - at least it did not blow up like last time but now the commit looks bigger than it is. #rebaseHatesMe


cleargif:
will redo PR

PR #531, opened 2015-01-04
Updates the JADU XSD

dan-palmer:
- [x] Updates the JADU XSD(0.19 -> 0.20)
- [x] Uses pounds not pence for fee amount.
- [x] Uses claim reference for the URN.
- [x] Adopts new address changes according to XSD change.
- [x] Improves E2E test for claim xml generation.
- [x] Removes some duplication in FactoryGril Claim traits


spec/factories.rb
```diff
@@ -39,8 +41,14 @@
       state        'created'
     end
 
+    trait :single_claimant do
+      without_additional_claimants_csv
+      after(:create) { |claim| claim.secondary_claimants.clear }
```
stevegraham: Does clear destroy the collection or null out foreign keys? Prob fine if the tests only only use instances returned by factory girl.

PR #532, opened 2015-01-05
Front end/ga virtual page views

cleargif:
Needs a review by @Demwunz and @TrentDSD


cleargif:
Redone this and will create new PR.

Issue #518, opened 2014-12-22
Maintenance page


cleargif:
Create a maintenance page that can be copied to separate environment

PR #534, opened 2015-01-06
Static service unavailable page

cleargif:
All files required for this page is in the folder. It does not have any assets outside of the
`/service-unavailable/ ` folder.

Tested on: IE 7, 8, 9, 10, 11, FF, Safari, Chrome


cleargif:
fixes #518

dan-palmer:
:hand: We don't need to duplicate all the assets here. The maintenance page just needs to have the correct paths for assets(css, images) & also for links.

cleargif:
Closing. Feature will be added in a new PR.

PR #535, opened 2015-01-06
Mount application under subdirectory (WIP)

stevegraham:
- Scope all routes under apply
 - Refactor session controller for cleaner routing
 - Actually log people out when they log out, i.e. clear session (WIP needs test)


stevegraham:
@Demwunz changed the sign out link to a button because it's doing a DELETE to the user sessions controller. Needs to be styled like a link if you want to keep the look and feel the same.

PR #536, opened 2015-01-06
Fixes filenames for claim attachments

dan-palmer:
- [x] make all the changes

### DIFF
--- a/app/services/carrierwave_filename.rb
+++ b/app/services/carrierwave_filename.rb
@@ -1,0 +1,7 @@
+module CarrierwaveFilename
+  def for(attachment)
+    attachment.file.path.split('/').last if attachment.file
+  end
+
+  extend self
+end
--- a/config/initializers/renderers.rb
+++ b/config/initializers/renderers.rb
@@ -1,5 +1,0 @@
-ActionController::Renderers.add :pdf do |obj, options|
-  filename = options[:filename] || 'data'
-  str = obj.respond_to?(:to_pdf) ? obj.to_pdf : obj.to_s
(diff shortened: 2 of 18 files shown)


config/initializers/renderers.rb
```diff
@@ -1,5 +0,0 @@
-ActionController::Renderers.add :pdf do |obj, options|
```
dan-palmer: :door: :wave: :walking:

dan-palmer:
Nah.. on second thoughts don't like this approach.

config/initializers/renderers.rb
stevegraham: See ya!

spec/models/claim_spec.rb
```diff
@@ -239,39 +241,38 @@
...
-      expect(subject.attachments.size).to eq 1
+      expect(subject.attachments.size).to eq 2
     end
   end
 
   describe "#generate_pdf!" do
     context "pdf has already been generated" do
       it "does not attempt to generate another" do
-        allow(subject).to receive(:pdf_blank?).and_return false
+        subject = create :claim, :with_pdf
```
stevegraham: Vas?

app/services/carrierwave_filename.rb
```diff
@@ -0,0 +1,7 @@
+module CarrierwaveFilename
+  def for(attachment)
+    attachment.file.path.split('/').last rescue nil
```
stevegraham: Would it be better to use a nil guard rather than swallowing exceptions.

dan-palmer:
@stevegraham removed that check on the pdf, didn't make much sense it being there & switched the nil munching out too.

PR #542, opened 2015-01-08
Mock AWS S3 in dev & test

dan-palmer:
- [x] Stub S3
- [x] Rejig tests after adopting such mockings
- [x] Fix file removals
- [x] Ensure files are actually deleted from S3
- [x] Fix all filenames

~~- [ ] Check carrierwave default store_dir path. Base uploader might not be required.~~

### DIFF
--- a/.env
+++ b/.env
@@ -6,3 +6,7 @@ EPDQ_SECRET_IN="m/mcr\\q]QKE'Pmal;ke2"
 EPDQ_SECRET_OUT="sEtw3k4m.a xl23lmw;la"
 
 JADU_API="https://etapi.employmenttribunals.service.gov.uk/1/"
+
+S3_UPLOAD_BUCKET="buckoff"
+AWS_ACCESS_KEY_ID="BADONDE"
+AWS_SECRET_ACCESS_KEY="brianbrianbrian"
--- a/app/forms/additional_claimants_upload_form.rb
+++ b/app/forms/additional_claimants_upload_form.rb
@@ -5,17 +5,20 @@ class AdditionalClaimantsUploadForm < Form
   attribute :remove_additional_claimants_csv, Boolean
 
   delegate :additional_claimants_csv_cache, :additional_claimants_csv_cache=,
-    :additional_claimants_csv_record_count=, :remove_additional_claimants_csv!,
-    :additional_claimants_csv_file, :reset_additional_claimants_count!, to: :target
+  :additional_claimants_csv_record_count=, :remove_additional_claimants_csv!,
+  :additional_claimants_csv_file, to: :target
 
-  before_validation :remove_csv!, unless: :has_additional_claimants
+  before_validation :remove_additional_claimants_csv!, if: :file_removal?
 
   with_options if: :has_additional_claimants do |form|
-    form.validates :additional_claimants_csv, presence: true
+    form.validates :additional_claimants_csv, presence: true,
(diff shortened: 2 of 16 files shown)


dan-palmer:
:ship:

stevegraham:
Looks good. Do we need a regression test? What would that look like? IMO the best thing would be to use the fog-mock back end in test. That would catch regressions wouldn't it?

dan-palmer:
Such agreements, will have a look into this today.

spec/features/create_claim_applications_spec.rb
```diff
@@ -214,10 +214,9 @@
     scenario 'Downloading the PDF if available' do
       complete_a_claim seeking_remissions: true
       click_button 'Submit claim'
-      click_link 'Save a copy'
 
-      expect(page.response_headers['Content-Type']).to eq "application/pdf"
```
dan-palmer: This test is :hankey:

dan-palmer: Had forgotten about this but the basic webkit driver with capybara doesn't like hitting external urls. So I moved those examples into a dedicated pdfscontroller spec.

Pdfs are tested poorly :cry: so asserting the contents is pretty important here & given we are going to be replacing this component I didn't want to do anything too funky.

app/forms/additional_claimants_upload_form.rb
```diff
@@ -15,7 +15,7 @@ class AdditionalClaimantsUploadForm < Form
     form.validates :additional_claimants_csv, content_type: {
       in: %w<text/csv text/plain>,
       message: I18n.t('errors.messages.csv')
-    }
+    }, unless: :remove_additional_claimants_csv
```
dan-palmer: The content validator won't work on the second run because the file isn't in the cache(local) & so calling path on the file object won't work but we can skip this if we are removing the file.

app/uploaders/base_uploader.rb
```diff
@@ -0,0 +1,9 @@
+class BaseUploader < CarrierWave::Uploader::Base
+  def store_dir
+    if Rails.env.test?
```
dan-palmer: This is now redundant :ghost:

dan-palmer:
With this approach the files are stored in memory which could be a bit troublesome in the dev environment. 

If you ever want to jump into a console session and get fiddling none of the files will be available on the models.

.env
```diff
@@ -6,3 +6,7 @@ EPDQ_SECRET_IN="m/mcr\\q]QKE'Pmal;ke2"
 EPDQ_SECRET_OUT="sEtw3k4m.a xl23lmw;la"
 
 JADU_API="https://etapi.employmenttribunals.service.gov.uk/1/"
+
+S3_UPLOAD_BUCKET="buckoff"
```
stevegraham: "BOOOOH!"

dan-palmer: These can be moved over to the .env.test when #554 is merged. :meat_on_bone:

app/validators/content_type_validator.rb
```diff
@@ -1,9 +1,16 @@
 class ContentTypeValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, _value)
     attachment = record.send(attribute)
+
+    return if already_attached_to_model?(attachment)
```
stevegraham: We'll do this with dirty tracking in a subsequent PR.

PR #543, opened 2015-01-08
Fix preselecting yes on additional claimants page

stevegraham:


app/models/claim.rb
```diff
@@ -68,9 +70,8 @@ def claimant_count
   end
 
   def remove_additional_claimants_csv!
-    update_attribute(:additional_claimants_csv_record_count, 0)
+    update_columns(additional_claimants_csv_record_count: 0, additional_claimants_csv: nil)
```
dan-palmer: i like this.. we don't have to run through the callbacks for the model again.

app/forms/collection_form.rb
```diff
@@ -41,7 +41,10 @@ def build_child
 
   def save
     if valid?
-      run_callbacks(:save) { collection.all?(&:save) }
+      run_callbacks(:save) do
```
dan-palmer: Does run callbacks treat whatever happens in the block as an atomic action and wrap it in a transaction?

spec/models/claim_spec.rb
```diff
@@ -55,7 +55,9 @@
 
     describe 'after_add on secondary_claimants' do
```
dan-palmer: after_add has been switched for a before_update cb

app/models/claim.rb
```diff
@@ -50,6 +49,9 @@ class Claim < ActiveRecord::Base
   before_update -> { secondary_claimants.destroy_all },
     if: :additional_claimants_csv_changed?
 
+  before_update :remove_additional_claimants_csv!,
+    if: -> (c) { c.secondary_claimants.any? }
```
dan-palmer: I think the lambda will be executed in the context of the claim so you can get away with passing any args

stevegraham: you're right although it's a lambda so it would still need an arity of 1 because it's called with the claim. could change it to:

```ruby
before_update :remove_additional_claimants_csv!,
  if: -> (_) { secondary_claimants.any? }
```

?

stevegraham: It looks like the arity is checked. I remember getting `ArgumentError`s earlier in the project doing this. Gonna change it to this 

```ruby
before_update :remove_additional_claimants_csv!,
  if: -> { secondary_claimants.any? }
```

spec/models/claim_spec.rb
stevegraham: léléléléléléléléléwl 🙈

app/forms/collection_form.rb
stevegraham: Nope it's not in a transaction. Do you think it should be. There are no validations on the model layer so it should never be false. I guess an exception might be raised in one of them if the DB died maybe? It's just a convenient way of returning true. Do you think we need to make it atomic?

app/models/claim.rb
stevegraham: *flutters eyelashes*

app/forms/collection_form.rb
dan-palmer: I think it makes sense to say its all or nothing however this is only a problem if the app has some sort of DB issue.

Speaking of DB problems.. http://www.postgresql.org/docs/9.4/static/runtime-config-autovacuum.html is something Ive seen cause issues before but I don't think we'll be doing enough traffic for this to be of a concern.

stevegraham: gonna wrap in transactions but in another PR.

dan-palmer: oki doki :+1:

PR #546, opened 2015-01-12
Improves the ACAS examples for the Respondent in the claim XML feature

dan-palmer:
Extracted from PR: #545


dan-palmer:
Failed on an unrelated spec. Assertion should probs use match_array unless order is important.

PR #556, opened 2015-01-14
Fazal Shnazzle Dates

dan-palmer:
- [ ] change gds date fields to have extra attribute and change type from text to number


Demwunz:
input should be `type="number"`. I also need to be able to add additional `pattern` attributes

dan-palmer:
@Demwunz If we use the number type it changes the desktop behaviour of the input fields. Rather than stopping a user only putting in 2 or 4 digits they can type as many as they want.

Also unless we pass a pattern like `pattern='\d*'` mobile safari will give you the keyboard with numbers and misc characters:

screen shot 2015-01-18 at 18 56 20

If we want the number keypad to show and keep the behaviour(honour only allowing 2 or 4 characters in desktop browsers) we could use `type="tel"`.

Made a wee fiddle for ya http://jsfiddle.net/o8tk1zk8/ :sparkles: :gun: 

ios simulator screen shot 18 jan 2015 18 59 55

dan-palmer:
:eyes: :up: cc @cleargif

dan-palmer:
see #565

dan-palmer:
#565 has been merged, such close?

PR #557, opened 2015-01-14
Ui/mobile menu manager

dan-palmer:


app/controllers/application_controller.rb
```diff
@@ -55,5 +55,21 @@ def claim_path_for(page, options = {})
     send "claim_#{page}_path".underscore, options
   end
 
-  helper_method :claim, :claim_path_for
+  def show_signout?
```
dan-palmer: + These methods should be hoisted up into a concern so that they can be used in the markdown controllers too. (Cookie spec is failing for this reason).
+ The views which previously took an options hash with the hide information are now irrelevant with this approach and need to be tidied up.
+ This approach changes how the removal of the sign out on the application-number page would have been implemented & should have been discussed with me & @stevegraham first.

app/controllers/application_controller.rb
```diff
@@ -55,5 +55,21 @@ def claim_path_for(page, options = {})
...
+    !@hide_signout
+  end
+
+  def hide_signout
+    @hide_signout = true
+  end
+
+   def show_mobile_nav?
```
dan-palmer: indentation :+1:

dan-palmer:
@cleargif @Demwunz 

One of you guys will have to rebase this beast. Too many changes in here I'm not sure on.

Demwunz:
I'll do it

Demwunz:
replaced with new branch ui/mobile-menu

PR #569, opened 2015-01-20
Static error pages

stevegraham:
Error pages for 404, 422, 500, and 503 HTTP errors

URLs:
http://0.0.0.0:3000/404.html
http://0.0.0.0:3000/422.html
http://0.0.0.0:3000/500.html
http://0.0.0.0:3000/503.html

@Demwunz There is an issue with the footer not being on the bottom of the page. Can you take a look? Also maybe we can remove the references to the JS too for these pages?

### DIFF
--- a/public/404.html
+++ b/public/404.html
@@ -123,13 +123,15 @@
   <div class="container">
     <header class="main-header">
       <h1>
-        The page you were looking for doesn't exist.
+        The page you are looking for does not exist or is temporarily unavailable.
       </h1>
     </header>
     <div class="main-content">
       <div class="static-content">
         <article id="service_suspended">
-          <p>You may have mistyped the address or the page may have moved.</p>
+          <div class="panel-indent">
+            <p>Please try retyping the address or go back to the previous page</p>
+          </div>
         </article>
       </div>
     </div>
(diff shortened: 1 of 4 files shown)


tomwm:
@stevegraham obviosuly i cant see the loal urls. Can i check you used the copy on here for the 500's
https://docs.google.com/a/digital.justice.gov.uk/document/d/1aN0dT1lPZFCdID_7DRCe5LCFwNpq3rOvX6mYPPo3DVM/edit for

stevegraham:
I didn't use that copy but @Demwunz can change it when he's going over it.

tomwm:
cool, make it so @Demwunz

Demwunz:
!(make it so)[http://amazingr.com/make-it-so.gif]

tomwm:
http://media3.giphy.com/media/MMpZTGSJn2eHu/200.gif

Demwunz:
@tomwm I've added a link on one page, but it all depends on how long the parallel service will be running for. Moreover, if they start an application in the other service and return, they will have committed to a system that will ultimately be phased out. 

There is also a danger that any 5xx 4xx error could just be a temporary glitch, by making them wait a few minutes or reloading would be a more suitable path of action.

stevegraham:
Thanks @Demwunz

PR #570, opened 2015-01-20
Replace non alpha numeric characters in filenames

dan-palmer:
Adds a rather sickening regex to replace characters in filenames that are not alphanumeric


app/services/carrierwave_filename.rb
```diff
@@ -2,7 +2,7 @@ module CarrierwaveFilename
   def for(attachment, underscore: false)
     file = attachment.file && attachment.file.path.split('/').last
 
-    underscore ? file.tr('-', '_') : file
+    underscore ? file.gsub(/[^a-zA-Z0-9](?=[[:alnum:]]*\.)/, '_') : file
```
dan-palmer: :sob:

stevegraham: Sanitisation could be hoisted into a private method with a descriptive name. It's no longer immediately clear what is happening here due to the relative complexity of the regex. Do you think it would help?

dan-palmer: Yeah totes agree. Sorry missed this comment, was considering another way to achieve the same result that conveys our intent a little better.

app/services/carrierwave_filename.rb
```diff
@@ -1,9 +1,18 @@
-module CarrierwaveFilename
```
stevegraham: The Module+extend self syntax will still work for private methods, no?

dan-palmer: lolz, I thought thats what I'd pushed - well spotted. was playing around with module_function and class separately.. & theres me adding a review label & not even looking over it myself. :boom: :gun: :skull:

PR #571, opened 2015-01-20
Fix AcasPresenter

stevegraham:
The mapping for security services had been changed and didn't match
up with the presenter. Fixed and added a regression test to catch
in the future.


dan-palmer:
nice one! :+1:

stevegraham:
@dan-palmer please review.

added a spec to catch such mismatches again and used fetch to blow up in future too making tracking the problem more straightforward if it somehow slips through.

PR #572, opened 2015-01-20
Ship sneakers.log and strip heartbeats

mattmb:
This will ship the sneakers log to logstash but remove the heartbeats which occur every 0.4s!


mattmb:
@dan-palmer @stevegraham happy for me to merge this, the failure seems unrelated.

PR #573, opened 2015-01-20
Setup deadletter queue

mattmb:
If there's an error, the message goes into the -retry queue, retries are managed by sneakers. If the max retries fail the message is sent to the -error queue.


mattmb:
@stevegraham @dan-palmer this is what got everything working when I was testing yesterday. I also had to set a policy in rabbit like this:
```
rabbitmqctl set_policy DLX "claim_submission" '{"dead-letter-exchange":"claim_submission-retry"}' --apply-to queues
```

stevegraham:
Do we need an exchange for the other jobs? e.g. mailers, fgr et al.

mattmb:
Yeah, sneakers seems to create everything except the dead letter policy. So I've put some salt code in to do that for fgr,mailers etc.

stevegraham:
merge if you have verified it works and you're happy. thanks for finding the solution.

PR #580, opened 2015-01-21
Check memorable word is present before trying to authenticate

### LINKED ISSUE #579: 500's if memorable word was never set and somebody tries to return to the claim

dan-palmer:
fixes #579 

*shoulda matcher perfoms valid `authenticates` checks with the `have_secure_password` matcher hence no test added or changed in the claim spec


stevegraham:
:+1:

PR #581, opened 2015-01-21
Fix fee group reference request behaviour

stevegraham:
FGR requests should only be made by the primary respondent. The
request code was written before multiple repsondents was implemented.
This commit ensures FGR requests are only made by the primary
respondent.


dan-palmer:
lookin good :sunglasses:

PR #595, opened 2015-01-27
Sanitise filenames

dan-palmer:


app/services/carrierwave_filename.rb
```diff
@@ -1,17 +1,25 @@
 module CarrierwaveFilename
   def for(attachment, underscore: false)
-    file = attachment.file && attachment.file.path.split('/').last
+    file_path = attachment.file && attachment.file.path
```
stevegraham: `path` - we're not using a file backend anywhere are? if not, we should be good.

dan-palmer: Nope, its all fog storage so we should be good.

stevegraham: kewl

PR #597, opened 2015-01-28
Fix worker timeouts

mattmb:
@dan-palmer happy with this? I've set some sane defaults and made it possible to pass in ENV vars to overwrite.

### DIFF
--- a/app/services/jadu/api/request.rb
+++ b/app/services/jadu/api/request.rb
@@ -3,11 +3,16 @@ require 'multipart'
 module Jadu
   class API
     class Request
+      HTTP_TIMEOUT_OPTIONS = {
+        open_timeout: ENV.fetch('HTTP_OPEN_TIMEOUT_S').to_i,
+        read_timeout: ENV.fetch('HTTP_READ_TIMEOUT_S').to_i }
+
       def initialize(uri, params, host: nil, **options)
         @uri = uri
         @params = params
         @host = host || uri.hostname
-        @options = options.merge(use_ssl: @uri.scheme == 'https')
+        @options = options.merge(use_ssl: @uri.scheme == 'https').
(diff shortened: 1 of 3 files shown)


app/services/jadu/api/request.rb
```diff
@@ -1,4 +1,6 @@
 require 'multipart'
+HTTP_OPTIONS = { open_timeout: ENV.fetch('HTTP_OPEN_TIMEOUT_S').to_i,
```
dan-palmer: I'd do it like this:

```ruby
require 'multipart'

module Jadu
  class API
    class Request
      HTTP_TIMEOUT_OPTIONS = {
        open_timeout: ENV.fetch('HTTP_OPEN_TIMEOUT_S').to_i,
        read_timeout: ENV.fetch('HTTP_READ_TIMEOUT_S').to_i }
                 
      def initialize(uri, params, host: nil, **options)
        @uri = uri
        @params = params
        @host = host || uri.hostname
        @options = options.merge(HTTP_TIMEOUT_OPTIONS).
                     merge(use_ssl: @uri.scheme == 'https')
      end

      def perform
        req = Multipart::Post.new(@uri.path, *@params)
        req['Host'] = @host
        Net::HTTP.start(@uri.hostname, @uri.port, @options) { |http|
          http.request(req)
        }
      end
    end
  end
end
```

dan-palmer:
Will need to have the `HTTP_OPEN_TIMEOUT_S` & `HTTP_READ_TIMEOUT_S` variables added to the Travis environment.

Me & @stevegraham suggest the environment variables should be added through the travis-ci GUI.

mattmb:
Cool. I need to add them to dev, staging too. I'll get on that tomorrow morning, then we can merge.

PR #599, opened 2015-01-29
Refactor/claim model

dan-palmer:
I’m not over the moon with this approach as it feels as though things are being brushed under the carpet. However.. claim is our god object and as such is always going to end up a bit unwieldy.. & looking at it that way extracting some ancillary methods and data into modules seems to make sense. Was inspecting the claim model and thinking the most important things I want to see are methods that relate to prominent actions we are performing through the application process along with the relationships a claim has with other models which led to the changes in this branch. Anyways food for thought & we can treat this purely as throw away in the bin codez.

### DIFF
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -1,6 +1,6 @@
 class Claim < ActiveRecord::Base
   include MemorableWord
-  include ClaimAttachments
+  include CarrierwaveAttachments
   include BitmaskedComplaintsOutcomes
 
   has_one :primary_claimant,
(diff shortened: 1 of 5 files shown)


stevegraham:
IMO this is just to satisfy CodeClimate, it doesn't improve the code. The problem code climate has is with the complexity of the code outside of method bodies, e.g. it doesn't like the scoped associations because of the lambdas. I did some refactoring removing this kind of thing where we could and it brought Claim back up to a B

dan-palmer:
Yeah this approach is :hankey: was kind of curious to see what score CC would give it though. The only module extraction on it that really makes sense is MemorableWord, rest is as you say just to satisfy CC.

PR #600, opened 2015-01-29
Ensure logs are created with container

mattmb:


docker/rails/logstash-conf.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash
+for i in /rails/log/unicorn.log /rails/log/production.log /rails/log/logstash_production.json /rails/log/sneakers.log;
```
dan-palmer: How abouts..

```bash
for i in unicorn.log production.log logstash_production.json sneakers.log;
    do touch /rails/log/$i
end
```

dan-palmer: @mattmb How abouts..

```bash
for file in unicorn.log production.log logstash_production.json sneakers.log;
    do touch /rails/log/$file
done
```

mattmb: yeah, I was lazy and copy pasting

PR #601, opened 2015-01-30
Admin dashboard for inflight metadata inspection

stevegraham:
Couple of things:

- There are no user accounts, anyone on the VPN can access this page. It's a 404 if they're not on the VPM.
- Pending the addition of user accounts the ability to change or otherwise interfere with claims is disabled.
- Similarly it is not possible to view any claimant or claim information, just metadata.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -53,6 +53,7 @@ end
 
 gem 'webmock', group: :test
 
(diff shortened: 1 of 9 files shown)


dan-palmer:
Such cool.. so nice! Good job @stevegraham :+1:

PR #608, opened 2015-01-30
Simplify some of the "complexity" outside of method bodies (CodeClimate)

stevegraham:


stevegraham:
This gets it back up to a B. I'd to get rid of the claimants and respondents associations not sure that's doable until we redo the pdf stuff.

PR #610, opened 2015-02-01
Hide save & complete later

dan-palmer:
Hide save and complete later button(claim#new, claim#show{application-number}, user_sessions{all}, feedback_form)

### DIFF
--- a/app/controllers/claims_controller.rb
+++ b/app/controllers/claims_controller.rb
@@ -2,6 +2,7 @@ class ClaimsController < ApplicationController
   redispatch_request unless: :created?, only: %i<show update>
 
   before_action :check_session_expiry, only: %i<show update>
+  before_action :hide_signout, only: %i<new show>
 
   def new
     @claim = Claim.new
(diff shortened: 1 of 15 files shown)


app/controllers/claims_controller.rb
```diff
@@ -49,5 +50,9 @@ def fee_calculation
     @fee_calculation ||= claim.fee_calculation
   end
 
+  def hide_signout
+    super if action_name == 'new' || page_manager.hide_signout?
```
stevegraham: What does this super to?

stevegraham: Ah ok, `SignoutVisibility`

stevegraham:
Not in love with the approach, i.e. seems to be a view concern but the controller is heavily involved. Off the top of my head:

```ruby
if SignoutVisibility.visible?
  # show button
end
```

`SignoutVisibility.visible?` would have access to the request (if it's not possible to have that by including helpers than it could be injected at call time) and it could compare the current path with a list of paths where signout links should be hidden. Thoughts?

PR #613, opened 2015-02-02
Render errors on base

stevegraham:
At this point we only use this on the user sessions page. The behaviour
is if there are errors on base show the errors in the notification
header, else show the fix errors below message.


stevegraham:
Looks like this
lel

Demwunz:
@stevegraham check last commit, removed excess whitespace.

stevegraham:
@demwunz thank you sir

Demwunz:
@stevegraham  no problemo

PR #614, opened 2015-02-02
Extract show signout into view helper

dan-palmer:


dan-palmer:
Needs its history fixing up

PR #616, opened 2015-02-04
Allow pdf to be generated when payment is required

dan-palmer:
- [x] Perform pdf generation asynchronously when payment is required(so we can hit up the pdf link on the payment page)
- [x] Ensure PDF is only generated once
- [x] Exchange config on RMQ @mattmb

### DIFF
--- a/app/jobs/pdf_generation_job.rb
+++ b/app/jobs/pdf_generation_job.rb
@@ -1,0 +1,7 @@
+class PdfGenerationJob < ActiveJob::Base
+  queue_as :pdf_generation
+
+  def perform(claim)
+    claim.generate_pdf!
+  end
+end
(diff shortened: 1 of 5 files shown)


mattmb:
Is this to catch an edge case? Or do you want to display the PDF to the user before the payment?

dan-palmer:
@mattmb PDF link needs to be shown on the submission page and on the (optional)payment page

mattmb:
OK :+1: I'll add the extra stuff to rabbit

dan-palmer:
@mattmb thank you sir :tophat: :smile:

mattmb:
Done in dev and staging, will push to prod if we don't have any problems.

dan-palmer:
Regarding the GUI side of things.. A link to the PDF will be included in #617 (it has a few other copy amends in it as well)

stevegraham:
Potential race condition, i.e. User clicking link before PDF is ready. What happens there?

PR #631, opened 2015-02-17
Fix up respondent addresses implemenation

stevegraham:


spec/features/claim_xml_generation_spec.rb
```diff
@@ -264,7 +264,7 @@ def xpath_collection(path)
       end
       it 'has Checksum(s)' do
         expect(xpath_collection("//Files/File/Checksum")).
-          to match_array %w<ee7d09ca06cab35f40f4a6b6d76704a7 58d5af93e8ee5b89e93eb13b750f8301 2e8c57f80b37129030867a4b003c6629>
+          to match_array %w<ee7d09ca06cab35f40f4a6b6d76704a7 58d5af93e8ee5b89e93eb13b750f8301 236ea37419ebc7e70a1a27ea7e43508e>
```
stevegraham: Not sure why the checksum has changed. I did a visual inspection of the PDF and both addresses were present and correct.

app/models/respondent.rb
```diff
@@ -11,17 +11,11 @@ class Respondent < ActiveRecord::Base
   before_save :enqueue_fee_group_reference_request, if: :needs_fee_group_reference?
 
   def address
-    addresses.first
+    addresses.detect(&:primary?) || addresses.build(primary: true)
```
stevegraham: Would have preferred to use `find_or_initialize_by` but these query methods don't consider unpersisted members of the collection, i.e. they just make a DB query.

dan-palmer:
such nice. :ship:

PR #633, opened 2015-02-18
Strip extraneous new lines from the PDF free text fields

stevegraham:


app/presenters/pdf_form/claim_presenter.rb
```diff
@@ -14,8 +14,28 @@ def to_h
     presenters.each_with_object({}) { |p, o| o.update p.to_h }
   end
 
+  def other_outcome
```
stevegraham: code climate may 💩

dan-palmer:
:+1:

PR #635, opened 2015-02-20
Blowing whistles is pricey

stevegraham:
Whistleblowing claims should attract the higher fee.

Renames Claim#alleges_discrimination_or_unfair_dismissal? to
Claim#attracts_higher_fee? and updates the logic of this method to
reflect the above.


dan-palmer:
:+1:

PR #637, opened 2015-02-20
Styling for Barclays epdq

fordie:
I've wrapped a 'payment-zone' class round the whole lot, just so we don't run into any inheritance issues, not that that's likely.


fordie:
payment_styled2
payment_styled1

Styled pages. These do have the MOJ header & footer, but I couldn't fit it in the screen grab

PR #638, opened 2015-02-20
.notice flash message styling

fordie:
changed colours & font size of .notice flash messages
make .notice flash message fade out after 10 seconds


fordie:
sorted out indentation.

I'm in two minds about auto-dismissing flash messages. This is only on .notice flash messages, so anything more serious would persist.

This is more of a nice to have than a must have behaviour, so treated it as progressive enhancement, so no. I haven't worried about polyfills.

If these colours are going to be used elsewhere then we should refactor to give them nice variable names instead of the hex values

PR #640, opened 2015-02-20
Revert "Adds representative type to XML - SUCH AVOID MERGING"

stevegraham:
Reverts ministryofjustice/atet#630


stevegraham:
Reverting per @helenmott's request

PR #648, opened 2015-02-26
Admin panel PDF generation & PDF downloads.

dan-palmer:
Feature to stop such PM's & DM's hassling the devs.. :laughing:

### DIFF
--- a/app/admin/claim.rb
+++ b/app/admin/claim.rb
@@ -26,6 +26,21 @@ ActiveAdmin.register Claim do
     column(:state) { |claim| claim.state.humanize }
   end
 
+  member_action :generate_pdf, method: :post do
+    PdfGenerationJob.perform_later resource
+    redirect_to :back, notice: 'Generating a new PDF'
+  end
+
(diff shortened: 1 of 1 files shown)


dan-palmer:
screen shot 2015-02-26 at 18 31 41

tomwm:
this is cool. if this goes in do we want to think bout something better than basic auth or will that be enough for now? @mattmb @stevegraham @TrentDSD @helenmott

mattmb:
Love how little code this took :+1: 

@dan-palmer how difficult would it be to separate the active admin stuff into a separate app so that we could run it in a different container to the main app?

stevegraham:
That's the advantage of picking the right tools and producing well designed, reusable code. Something we've managed to do a good job of on this project.

+1 admin panel being another app. Only lives in the main app for expediency.

dan-palmer:
@mattmb hmm perhaps we just submodule etfees repo and have an etfees-admin repo dedicated to this. Thoughts? @stevegraham @mattmb

PR #649, opened 2015-03-03
Replaces quotation marks that are otherwise classed as undefined

dan-palmer:


dan-palmer:
Would like a little longer to investigate this. It solves the immediate problem at hand but I'm not convinced this is correct.. I mean how does this scale? It just doesn't.

PR #651, opened 2015-03-05
Check the claim type in pdf & populate the number of claimants field

dan-palmer:
screen shot 2015-03-05 at 15 07 37
screen shot 2015-03-05 at 15 06 53

### DIFF
--- a/app/presenters/pdf_form/claim_presenter.rb
+++ b/app/presenters/pdf_form/claim_presenter.rb
@@ -41,8 +41,22 @@ class PdfForm::ClaimPresenter < PdfForm::BaseDelegator
     claim_types.any? { |type| pay_claims.include? type }
   end
 
+  def type_of_claim
+    if has_multiple_claimants?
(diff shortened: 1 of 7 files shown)


app/presenters/pdf_form/claim_presenter.rb
```diff
@@ -41,8 +41,22 @@ def owed?
     claim_types.any? { |type| pay_claims.include? type }
   end
 
+  def type_of_claim
+    if claimant_count > 1
+      'a claimon behalf of more than one person'
```
stevegraham: lél this pdf is crap

stevegraham:
nice work

app/presenters/pdf_form/claim_presenter.rb
dan-palmer: hahaha such!

dan-palmer:
Added a query method to claims public interface (shouldn't have forced pushed though, please forgive)

PR #661, opened 2015-03-30
Shorter strings for pdf fields.

mattmb:
This reduces the maximum number of characters for claim details and additional information. This should prevent truncation of these fields in the fixed space of the PDF boxes.

### DIFF
--- a/app/forms/additional_information_form.rb
+++ b/app/forms/additional_information_form.rb
@@ -5,7 +5,7 @@ class AdditionalInformationForm < Form
   before_validation :reset_miscellaneous_information!,
     unless: :has_miscellaneous_information?
 
-  validates :miscellaneous_information, length: { maximum: 5000 }
+  validates :miscellaneous_information, length: { maximum: 2500 }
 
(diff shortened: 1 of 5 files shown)


dotemacs:
This one is not too bad. Nothing programmatic here. 

* But have you tested the PDF form fields to see what is their max capacity before making this change? 
* What I'm asking is what made you decide to cut the maximum capacity of 5000 to 2500?

mattmb:
Yeah, it's kind of difficult to test because it varies depending on things like number of paragraphs. In the long run we want to make it dynamic.

tomwm:
All paragraphs a reduced to 1 line break only i believe (Stevie did this before). So surely we can work out the maximum number of line breaks possible?

dotemacs:
@mattmb what are your thoughts on what @tomwm mentioned above :question:

mattmb:
We have agreed to tackle it in a separate PR.

tomwm:
what @mattmb said

PR #662, opened 2015-03-30
Email validation

mattmb:
This adds email validation to the user's and representitives' email addresses. This is mostly to ensure that no spaces are entered in the email address as it causes ActionMailer to blow up.


dotemacs:
Looks fine to me, will merge it now.

Issue #679, opened 2015-03-31
Change the payment button for partial remission


milror00:
It should be classified as full payment

dan-palmer:
@milror00 is this issue solved in #714

cc @tomwm

tomwm:
yes. fixed by #714

PR #714, opened 2015-04-20
Submit Claim Button Text Fix For Group Claims

dan-palmer:
Prior to this PR - a group with partial remission showed the button text that a full remission claim would have. 

The cause of this was due to the method for presenting the button text checked only remission had been applied - it now checks the remission count instead.


tomwm:
fixes #679

PR #723, opened 2015-04-28
fixes #699

### LINKED ISSUE #699: Change copy around remission application
On 'your fee' page we need to add text. The lawyers want what is below but we should massage it a little

“Having read the guidance do you submit an application for remission?”
 It should then say next to the yes box –
"By ticking this box you have made a fee remission application.  If you do not complete the requisite form and provide supporting evidence to the Tribunal within seven days your fee remission application will be rejected”.

Demwunz:
fixes #699


dan-palmer:
Just checked this out. Looks good. Nice one @Demwunz

PR #730, opened 2015-05-08
Guide Changes + Back To Form From Guide Helper

### LINKED ISSUE #692: Guide Changes
Make changes to guide to bring it more in line with Plain English

https://docs.google.com/a/digital.justice.gov.uk/document/d/1TYCUjlZNuqIpS3a_iTE5HA_UqkHom7hcyDK-c2nT8mI/edit

### LINKED ISSUE #720: Provide route back to the form from the Guide
Users concerned when they were trying to return from the guide as there is no back button but it had opened in the same tab. Some users do not know how to return to the form. Some are concerned that they may lose what they have already input if they use the browser back button.

Demwunz:
Fixes #720 
Fixes #692 

lord @michaelelwin decrees the changes are fit and must go live.

seal of approval

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -426,7 +426,7 @@ en:
       writing_your_claim_statement: Writing your claim statement
       aside:
         gethelp_header: Support
-        exit_guide: Back to review page
+        exit_guide: Return to form
         contact_us: Contact us
         contact_us_link: http://www.justice.gov.uk/contacts/hmcts/tribunals/employment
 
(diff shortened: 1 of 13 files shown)


dan-palmer:
@Demwunz this needs some work to handle the link form the guide back to the last form page the user was on right?

config/locales/en.yml
```diff
@@ -366,6 +366,7 @@ en:
       gethelp_header: Support
       actions_header: Your claim
       read_guide: Guide
+      exit_guide: Back to form
```
dan-palmer: where is this used?.. same applies to Line 379

looks like they shouldn't be here

@Demwunz

Demwunz: @dan-palmer this is for the link back to the form - see line 20

Issue #731, opened 2015-05-11
reconfigure the workers so they aren't going mental


tomwm:

mattmb:
https://github.com/ministryofjustice/atet/pull/736 @tomwm

tomwm:
thanks!

PR #735, opened 2015-05-11
Make workers configurable

mattmb:
@dan-palmer just a small change here so that I can set the number of workers through env vars.

To solve #731

### DIFF
--- a/config/initializers/amqp.rb
+++ b/config/initializers/amqp.rb
@@ -4,12 +4,13 @@ if Rails.env.production?
   Sneakers.configure \
     amqp:    "amqp://#{ENV.fetch('RABBIT_USER')}:#{ENV.fetch('RABBIT_PASSWORD')}@#{ENV.fetch('RABBIT_HOST')}:#{ENV.fetch('RABBIT_PORT')}",
     vhost:   ENV.fetch('RABBIT_VHOST'),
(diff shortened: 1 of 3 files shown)


config/initializers/amqp.rb
```diff
@@ -4,6 +4,7 @@
   Sneakers.configure \
     amqp:    "amqp://#{ENV.fetch('RABBIT_USER')}:#{ENV.fetch('RABBIT_PASSWORD')}@#{ENV.fetch('RABBIT_HOST')}:#{ENV.fetch('RABBIT_PORT')}",
     vhost:   ENV.fetch('RABBIT_VHOST'),
+    workers: ENV.fetch('SNEAKERS_WORKERS').to_i,
```
dan-palmer: I'd suggest doing a parse to`Integer()` here too. 
`Integer ENV.fetch(:my_var)`

If you send `:to_i` to `nil` or some garbage string you'll get a zero back.. not sure how sneakers will handle that & if it will throw an error so its a bit misleading. 

Anything you expect should be a number is probz a good idea to parse with `Integer()`.

mattmb: I'm sold :smile: was wondering if there was any difference. Will change it now.

dan-palmer:
:ship:

PR #736, opened 2015-05-13
Make workers configurable

dan-palmer:
This makes it possible to define the number of worker processess for
sneakers and unicorn through environment variables.

Also move to parsing integers with the Integer class rather than .to_i
which is safer because it throws an exception if given a string as
input.


tomwm:
fixes #731

PR #738, opened 2015-05-13
Only Run Session Timeout Prompt On Claim Form Pages

### LINKED ISSUE #588: Frontend Session Prompt Runs On Every Page
Since we put everything under /apply/ the session prompt will be triggered on every page.

`moj.session-prompt.js`
```javascript
 var isClaimPath = function () {
    return !!location.pathname.match(/^\/apply\/.*/);
  };

  if (isClaimPath()) {
    sessionPrompt.init();
  }
```

dan-palmer:
Only run the session prompt on actual claim form pages.

The change to nesting all paths in the app under '/apply' caused the original implementation to fire up the session prompt on all pages.

fixes #588


app/views/layouts/application.html.slim
```diff
@@ -15,6 +15,10 @@
     javascript:
       gaTracker.init('#{ENV.fetch("GOOGLE_ANALYTICS_ID")}');
 
+  - if controller.controller_name == 'claims'
```
dan-palmer: one of the problems with this approach as it stands is that the review page would also require a session prompt and that is rendered through a different controller

dan-palmer: @Demwunz

dan-palmer:
will test this by checking sessionprompt.init() is called on the relevant pages.

PR #741, opened 2015-05-15
Strips whitespace from the primary claimants first & last name

### LINKED ISSUE #707: Strip whitespace form claimants first & last name
Eventually leads to whitespace in the pdf file name.

dan-palmer:
PDF filenames can no longer have spaces in them as first and last names values are now 'stripped' before being persisted.

Fixes #707


dan-palmer:
CC no like define_method :sob:

PR #747, opened 2015-05-19
Adds a presenter for confirmation email addresses on review page

### LINKED ISSUE #737: Direct all outbound correspondence to representative, if present.
If rep details are present, all outbound emails from the app should go to them and NOT the primary respondent

dan-palmer:
Presenter that returns a list of emails & determines the checked properties of the emails based on which emails are present in the claim.

Fixes #737

### DIFF
--- a/app/presenters/confirmation_email_addresses_presenter.rb
+++ b/app/presenters/confirmation_email_addresses_presenter.rb
@@ -1,0 +1,26 @@
+class ConfirmationEmailAddressesPresenter < Struct.new(:claim)
+  delegate :primary_claimant, :representative, :secondary_claimants, to: :claim
+
+  def self.email_addresses_for(claim)
+    new(claim).filter_email_addresses
+  end
+
+  def filter_email_addresses
+    [primary_claimant_email, representative_email].
+      reject { |email| email.first.blank? }
+  end
+
+  private
+
+  def primary_claimant_email
+    [primary_claimant.email_address, checked: tick_primary_claimant?]
+  end
+
+  def representative_email
+    [representative.try(:email_address), checked: true]
+  end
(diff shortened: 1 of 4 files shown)


app/presenters/confirmation_email_addresses_presenter.rb
```diff
@@ -0,0 +1,26 @@
...
+  def primary_claimant_email
+    [primary_claimant.email_address, checked: tick_primary_claimant?]
+  end
+
+  def representative_email
+    [representative.try(:email_address), checked: true]
+  end
+
+  def tick_primary_claimant?
+    !(secondary_claimants.any? && representative.try(:email_address).present?)
```
threedaymonk: I think it's clearer to rewrite the logic:

```ruby
secondary_claimants.none? || representative.try(:email_address).blank?
```

dan-palmer:
@threedaymonk Proposed logic change has been made, excellent suggestion. :+1:

app/presenters/confirmation_email_addresses_presenter.rb
```diff
@@ -0,0 +1,26 @@
+class ConfirmationEmailAddressesPresenter < Struct.new(:claim)
+  delegate :primary_claimant, :representative, :secondary_claimants, to: :claim
+
+  def self.for(claim)
```
threedaymonk: I'd be inclined to rename this to something like `email_addresses_for`, as it's not otherwise clear what it returns.

dan-palmer: Makes sense. Originally I had this as a service object(without a suffix) & felt a little more expressive regarding what it returns.

PR #755, opened 2015-05-28
Adds A Text Area Character Counter

Demwunz:


Demwunz:
@dan-palmer if you're happy, please pull this in

app/assets/javascripts/modules/moj.countable.js
```diff
@@ -0,0 +1,51 @@
...
+        return max;
+      } else if(remaining >= max) {
+        return '0 ';
+      } else {
+        return remaining;
+      }
+    }
+  };
+
+  if($('textarea').length){
```
dan-palmer: @Demwunz Think we can remove this if condition, if there aren't any textareas I think it will just return an empty list.

Demwunz: @dan-palmer thats the intention. it stops the rest of the script from running if textareas aren't present in the page

dan-palmer: but if it's an empty list it wouldn't initialise the counters

Demwunz: thats right

PR #762, opened 2015-06-03
Started & Completed Claims Counts As JSON

dan-palmer:


app/models/stats/claim_stats.rb
```diff
@@ -0,0 +1,11 @@
+module Stats
+  class ClaimStats < Claim
+    scope :started, -> { where.not(password_digest: nil, state: 'submitted').within_timeframe }
+    scope :started_count, -> { started.count }
```
threedaymonk: This feels like a bit of a subversion of `scope` - it works because `scope` is essentially the same as defining a class method, but I think it's a bit confusing in terms of intention: it returns something that can't be chained (if I've understood correctly).

app/models/stats/claim_stats.rb
```diff
@@ -0,0 +1,11 @@
...
+
+    scope :completed, -> { where(state: 'submitted').within_timeframe }
+    scope :completed_count, -> { completed.count }
+
+    scope :within_timeframe, -> { where('created_at > ?', 92.days.ago) }
```
threedaymonk: Why is 92 important here?

app/models/stats/claim_stats.rb
dan-palmer: Knowing that `scope` is just basically sugar for a class method I probably did this to be a little less verbose :mask: & also because its suffixed with count I thought the intention would be clear - not the case. I will change this, as it does break the expected return value of a `scope` which should always be a `Relation`. 

Thanks for pointing this out as although I did want to have a proper glance over this before assigning it I doubt I would have changed this.

app/models/stats/claim_stats.rb
TrentDSD: 92 days = 3 months which is requirement

dan-palmer: I think rather than assign a constant to the number of days here I'll just change the scope name. Apart from that as this data is required for a single user who know the time range already I'm not going to change the json output to give any indication of date ranges. I think if we need to offer data to more 'things' then we should readdress this then and think about the API a little more. This is also the reason why I bundled the two values in a single payload.

threedaymonk: @TrentDSD my subtext was that it should be a constant.

92 is an interesting number, though: it's not a multiple of 3; it's the maximum of all three-month sequences (30 + 31 + 31). I'd put that in the source.

threedaymonk:
Looks fine to me.

dan-palmer:
thanks @threedaymonk

PR #763, opened 2015-06-03
Log events when the status of a successful barclaycard callback is uncertain

dan-palmer:
Assigns an event to a claim for instances when barclaycard cannot guarantee at the time a user submits payment whether the payment has actually been authorised.

### DIFF
--- a/app/services/payment_gateway/response.rb
+++ b/app/services/payment_gateway/response.rb
@@ -1,7 +1,7 @@
-# Namespace for encapsulating Payment gateway related code
 module PaymentGateway
-  # Encapsulates ePDQ response related code
   class Response < Struct.new(:request)
+    SUCCESS_CODES = %w<5 51 9 91>.freeze
+
     def valid?
       EPDQ::Response.new(request.query_string).valid_signature?
     rescue RuntimeError
@@ -11,7 +11,7 @@ module PaymentGateway
     end
 
     def success?
-      %w<5 51 9 91>.include? params['STATUS']
+      SUCCESS_CODES.include? status
     end
 
     def amount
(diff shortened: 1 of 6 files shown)


app/services/payment_gateway/response.rb
```diff
@@ -11,7 +11,7 @@ def valid?
     end
 
     def success?
-      %w<5 51 9 91>.include? params['STATUS']
+      %w<5 51 9 91>.include? status
```
threedaymonk: I'd suggest pulling this out to a constant.

dan-palmer: Good spot :+1:

app/presenters/barclaycard_response_message.rb
```diff
@@ -0,0 +1,22 @@
...
+    'Status 92: Payment Uncertain - The customer should not retry the authorisation
+    process since the authorisation/payment might already have been accepted.'.squish.freeze
+
+  def for(status)
+    case status
+    when '52' then AUTHORISATION_NOT_KNOWN
+    when '92' then PAYMENT_UNCERTAIN
+    else "Status Code: #{status}"
+    end
+  end
```
threedaymonk: I think this entire class could be replaced with Rails's i18n support:

```ruby
I18n.t(status, scope: 'barclaycard.whatever', default: :generic, status: status)
```

With `en.yml` being something like:

```yaml
en:
  barclaycard:
    whatever:
      "52": "Status 52 …"
      "92": "Status 92 …"
      generic: "Status Code: %{status}"
```

This would also have the nice effect of not requiring multi-line string jiggery-pokery in Ruby.

dan-palmer:
@threedaymonk Have pulled out the presenter and used I18n as suggested.

threedaymonk:
All looks good to me now.

dan-palmer:
@threedaymonk :+1:

Issue #765, opened 2015-06-04
slim labels broken


Demwunz:
some slim labels in the form now have a background and an unwanted border

tomwm:
fixed by #768

PR #768, opened 2015-06-08
Slim Label Fix

Demwunz:
addresses an issue where slim labels were messed up due to a change elsewhere in sass templates.


tomwm:
@demwonz does that fix #765

Demwunz:
@tomwm that's the one!

PR #772, opened 2015-06-15
added abbreviations where necessary for screenreaders

Demwunz:


dan-palmer:
@Demwunz This adds some breaking changes to the texts. By default the I18n translations have html escaped. So what you'll see is something like this:

```html
<div>
    I'm some lovely <abbr title="text">TXT</abbr> hehe.
</div>
```
Will render: <b>I'm some lovely  hehe.</b>
Instead of: <b>I'm some lovely TXT hehe.</b>

All the keys in the language file will need to be appended with '_html' and the relevant key lookup in the slim template will need to be changed accordingly.

PR #777, opened 2015-07-16
Adds a healthcheck endpoint

dan-palmer:


dan-palmer:
History needs tidying up.

@mattmb can we do a build of this and kill things, check if it works.

Grab me and we can have a chat.

mattmb:
LGTM @dan-palmer @robmckinnon :+1: 

I tried a few obvious ways to break each service in staging and it responded as I'd expect. The caching also worked as I'd expect.

cc @KamalaHB

PR #779, opened 2015-07-22
Feature/cache healthcheck

robmckinnon:
* Cache healthcheck report for 30 seconds, with a 5 second ttl.
* Create `HealthcheckReport` class for storing static result data.

### DIFF
--- a/app/services/healthcheck.rb
+++ b/app/services/healthcheck.rb
@@ -1,34 +1,23 @@
-class Healthcheck
+module Healthcheck
   COMPONENTS = [
     HealthcheckComponent::RabbitMq,
     HealthcheckComponent::BarclaycardGateway,
     HealthcheckComponent::Sendgrid
   ].freeze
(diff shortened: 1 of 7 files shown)


app/services/healthcheck.rb
```diff
@@ -1,30 +1,22 @@
-class Healthcheck
+module Healthcheck
```
dan-palmer: Given its public interface is the `report` method which is always interacting with the cache, how about renaming this to `CachedHealthcheck`?

dan-palmer: Think its slightly more intention revealing form the controller level too.

robmckinnon: Feels like caching an implimentation detail that the controller doesn't need to know about?

If we leave it as `Healthcheck` and happen to take out caching later, then we won't need to rename the module or change the controller code.

PR #780, opened 2015-07-22
Improve sendgrid spec to test helo() called.

robmckinnon:


spec/services/healthcheck_component/sengrid_spec.rb
```diff
@@ -2,34 +2,27 @@
...
 
-  it_behaves_like 'a healthcheck component'
-
   subject { described_class.new }
 
-  context 'sengrid is available' do
-    let(:successful_smtp_response) do
-      instance_double Net::SMTP::Response, success?: true
-    end
+  let(:response_success) { true }
```
dan-palmer: I like the way the `response_status` has been scooped to the sendgrid is unavailable context, lets move this one into the sendgrid is available context.

dan-palmer:
Other than comment I made this is spot on! :+1:

spec/services/healthcheck_component/sengrid_spec.rb
robmckinnon: Ok, I'll move setting `response_status` inside the context for clarity. Means I'll move `it_behaves_like 'a healthcheck component'` into each context.

PR #788, opened 2015-09-09
Set sneakers to log to STDOUT

mattmb:
Get sneakers to log to STDOUT and then capture the log using the init
system. Trying this because otherwise sneakers rotates its own log file
and occassionally breaks the shipping.


dan-palmer:
@mattmb perhaps give it a whirl on staging to check da thingzzz?

otherwise.. LGTM

mattmb:
Cool thanks, have done already and it seems to work :smile: 

Will deploy and smoke test after merge too.

PR #791, opened 2015-10-06
Correct RepresentativeType spelling error

dan-palmer:
Trade Association was spelt incorrectly & as it is validated against the
schema it can cause claims to fail validation and not be sent through to
Jadu.


andyhd:
:+1:

tomwm:
LGTM

PR #794, opened 2015-10-28
Upgrade Ruby v2.2.3 & Rails v4.2.4

dan-palmer:


dan-palmer:
Tests appear to be taking twice as long. Need to investigate.

dan-palmer:
This is old.. we should be looking at bumping this to ruby 2.3.0 & latest rails 4.2.6

PR #799, opened 2015-11-23
WIP Moving to ruby 2 2 3

lukaszraczylo:


filipposc5:
We can probably close this now, as I 've pulled in the application building and container all together in #805  which results to a usable image.

PR #802, opened 2016-01-04
Add payment gateway task logging

filipposc5:
Adds logging to the payment gateway when the state changes from
available = true to false and vice versa. This should allow someone
to detect whether the gateway ever becomes unavailable
(spoiler alert: it does become unavailable often).

### DIFF
--- a/app/services/payment_gateway.rb
+++ b/app/services/payment_gateway.rb
@@ -6,13 +6,26 @@ module PaymentGateway
   # in those environments, but it's annoying in development
   @available = true
   MUTEX = Mutex.new
+  logger = Rails.logger
+  logger.info "Initialised PaymentGateway logger"
 
   TASK = PeriodicTask.new(every: 5.seconds, run_immediately: !Rails.env.test?) do
     begin
       result = HTTParty.get ENV.fetch('PAYMENT_GATEWAY_PING_ENDPOINT')
-      MUTEX.synchronize { @available = result.success? }
-    rescue SystemCallError
-      MUTEX.synchronize { @available = false }
+      MUTEX.synchronize {
+        if result.success? != @available
+          logger.info "PaymentGateway: State change from old #{@available} to #{result.success?}"
+        end
+        # leaving this outside of the 'if' because when changed the tests break
+        @available = result.success?
+      }
+    rescue SystemCallError => e
+      logger.error "PaymentGateway: SystemCallError - #{e.class} #{e.message} " +
+        " \n    " + e.backtrace.join("\n    ") +
+        "\nPaymentGateway: SystemCallError - gateway now unavailable (#{e.message})"
+      MUTEX.synchronize {
+        @available = false
+      }
     end
   end
 
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -114,6 +114,7 @@ class Claim < ActiveRecord::Base
   end
 
   def payment_applicable?
+    Rails.logger.info "Claim #{self.id} #{self.application_reference} payment_applicable gw: #{PaymentGateway.available?} fee_to_pay? #{fee_to_pay?} fee_group_reference #{fee_group_reference?} "
     PaymentGateway.available? && fee_to_pay? && fee_group_reference?
   end
 


filipposc5:
@dan-palmer  I have added this logging and while testing I can see this is happening: 

```
I, [2016-01-05T14:25:10.505458 #3681]  INFO -- : Initialised PaymentGateway logger
I, [2016-01-05T14:26:10.442287 #3696]  INFO -- : Initialised PaymentGateway logger
I, [2016-01-05T14:27:10.349913 #3710]  INFO -- : Initialised PaymentGateway logger
I, [2016-01-05T14:28:10.575225 #3724]  INFO -- : Initialised PaymentGateway logger
I, [2016-01-05T14:29:10.669607 #3738]  INFO -- : Initialised PaymentGateway logger
I, [2016-01-05T14:30:10.354164 #3752]  INFO -- : Initialised PaymentGateway logger
```

That also means every 1 minute this 'health check' resets to available = true. Any idea why this is happening?

filipposc5:
Tested the exception handling/logging as well, this now gives a nice error in logs: 
```
E, [2016-01-14T11:55:55.916556 #34] ERROR -- : PaymentGateway: SystemCallError - Errno::ETIMEDOUT Connection timed out - connect(2) for "10.0.1.1" port 443
    /usr/local/lib/ruby/2.2.0/net/http.rb:879:in `initialize'
    /usr/local/lib/ruby/2.2.0/net/http.rb:879:in `open'
    /usr/local/lib/ruby/2.2.0/net/http.rb:879:in `block in connect'
    /usr/local/lib/ruby/2.2.0/timeout.rb:74:in `timeout'
    /usr/local/lib/ruby/2.2.0/net/http.rb:878:in `connect'
    /usr/local/lib/ruby/2.2.0/net/http.rb:863:in `do_start'
    /usr/local/lib/ruby/2.2.0/net/http.rb:852:in `start'
    /usr/local/lib/ruby/2.2.0/net/http.rb:1375:in `request'
    /rails/vendor/bundle/ruby/2.2.0/gems/httparty-0.13.3/lib/httparty/request.rb:98:in `perform'
    /rails/vendor/bundle/ruby/2.2.0/gems/httparty-0.13.3/lib/httparty.rb:539:in `perform_request'
    /rails/vendor/bundle/ruby/2.2.0/gems/httparty-0.13.3/lib/httparty.rb:475:in `get'
    /rails/vendor/bundle/ruby/2.2.0/gems/httparty-0.13.3/lib/httparty.rb:577:in `get'
    /rails/app/services/payment_gateway.rb:14:in `block in <module:PaymentGateway>'
    /rails/app/services/periodic_task.rb:11:in `call'
    /rails/app/services/periodic_task.rb:11:in `block (2 levels) in initialize'
    /rails/app/services/periodic_task.rb:10:in `loop'
    /rails/app/services/periodic_task.rb:10:in `block in initialize'
PaymentGateway: SystemCallError - gateway now unavailable (Connection timed out - connect(2) for "10.0.1.1" port 443)
```

filipposc5:
The payment gateway logging initialising seems to happen because of the way the healthcheck component is structured (and is hit frequently by our monitoring tools). This may need to be revisited in the future.

app/services/payment_gateway.rb
```diff
@@ -6,13 +6,25 @@ module PaymentGateway
...
     begin
       result = HTTParty.get ENV.fetch('PAYMENT_GATEWAY_PING_ENDPOINT')
-      MUTEX.synchronize { @available = result.success? }
-    rescue SystemCallError
-      MUTEX.synchronize { @available = false }
+      MUTEX.synchronize {
+        if result.success? != @available
+          logger.info "PaymentGateway: State change from old #{@available} to #{result.success?}"
+        end
+        @available = result.success?
```
aldavidson: this line should be inside the if, shouldn't it? otherwise it's redundant

app/services/payment_gateway.rb
```diff
@@ -6,13 +6,25 @@ module PaymentGateway
...
+      }
+    rescue SystemCallError => e
+      MUTEX.synchronize {
+        logger.error "PaymentGateway: SystemCallError - #{e.class} #{e.message} " +
```
aldavidson: this log line doesn't need to be inside the mutex - try to lock the absolute minimum that you need to lock

aldavidson:
The logging additions look fine, apart from the two minor stylistic comments I made.

There is something about the way this module is using module-scope that makes me a bit nervous - shared state at module level can be v. tricky to manage correctly in a production environment, especially with a forking/pre-loading app server like unicorn. I wouldn't be surprised if this results in some unpredictable concurrency issues. However, that was there before this commit, so it's an existing issue and shouldn't stop this PR being accepted

PR #804, opened 2016-04-01
ET1 PDF update

dan-palmer:
The ET1 form has been update & in order for the application to be brought up to date the PDF template we use to inject content into needs to be replaced along with the syncing up of any respective changes made in the new PDF.

The new document removes an acas exemption reason which requires the same radio selection to be removed from the application.

With the removal of one of the acas exemption reasons we now have claims that people can return to(i.e not submitted to jadu yet) that have the old acas reason associated with one or more of the respondents. This shouldn't be a problem though as the save and complete forces a user to go through all of the form pages when the user returns. As the old reason is no longer in the the accepted reasons lists the respondents & additional respondents pages will fail validation, showing the user an inclusion error message prompting them to select a new reason.

Claims that have been marked as submitted or payment required needn't be concerned with this legacy change as the claim is locked out & besides Jadu would have already received the claim.


dotemacs:
Like I mentioned in person to @dan-palmer the code looks fine, it's mostly business logic here...

PR #805, opened 2016-04-07
Docker next gen

filipposc5:
First working version which brings up unicorn and allows priviledged / unpriviledged execution. 
Based on https://github.com/ministryofjustice/atet/pull/799 and previous work in ET. 

How to build/run instructions in  2594078100d58ff04d38ec983c3ff2d3ffe28ce4

**TODO suggestions**
1. Add a mechanism (env var? ENABLE_SNEAKERS perhaps) to more cleanly decide if we need sneakers workers or not, rather than checking if it is "production". 
2. Integrate with docker compose / adjust variables for database etc. 
3. Figure out what to do with assets (S3?)
4. Add any other services in init eg sneakers if env var from 1. is set, or logstash etc.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+docker-ng/rails/Dockerfile
+docker-ng/assets/Dockerfile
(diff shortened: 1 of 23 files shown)


lgarvey:
This is no longer needed.

PR #807, opened 2016-05-04
Fix for a old link

dotemacs:
The content was updated, but the old URL was left in.

This change fixes that.


leighmoney:
@dotemacs - looks good to me, thanks

PR #841, opened 2017-01-25
Making sure that at least one claim type is submited.

zaparka:
JIRA ticket: https://triadmoj.atlassian.net/browse/RST-12

As a public Employment Tribunal user, I want the system to make it mandatory to select at least one 'claim type', so I can comply with the Employment Tribunal Rules of Procedure for presenting a claim and the correct fee can be applied.

AC 1: The user cannot move to the next page without selecting at least one of the claim data types.
AC 2: Where a user proceeds without selecting a claim data type, system must present an error message.
AC3: Information from the form is stored in the Employment Tribunal database.
AC4: The correct fee has been applied to the ET application. Note: There are no changes required to the fee calculation.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,3 +1,3 @@
 --color
 --require spec_helper
---format documentation
+--format p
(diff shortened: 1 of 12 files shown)


zaparka:
@lgarvey please take a look.

PR #853, opened 2017-04-04
Feature/gems update

zaparka:
Updating gems marked as a security risk by Gemnasium.


leoapost:
I see that atet uses the deprecated [moj_template](https://github.com/ministryofjustice/atet/blob/master/app/views/layouts/application.html.slim#L41) which is [loading the jQuery](https://github.com/ministryofjustice/moj_template/blob/master/source/views/layouts/partials/_body_end.html.erb#L2) lib, so I wonder if we also need it in the Gemfile. 

In any case, are we sure bumping up jQuery's version will not break IE?

zaparka:
@leoapost yes that's the reason why I updated the uglifier only to certain version as you did on courtfinder.

PR #854, opened 2017-04-18
Adding direct csv DoB validation

zaparka:
This is here because when we sending data to Jadu it has to be in specific format. https://triadmoj.atlassian.net/browse/RST-201

### DIFF
--- a/app/services/additional_claimants_csv/model_builder.rb
+++ b/app/services/additional_claimants_csv/model_builder.rb
@@ -1,5 +1,4 @@
 class AdditionalClaimantsCsv::ModelBuilder
-
   ATTRIBUTES = %i<title first_name last_name date_of_birth address_building
     address_street address_locality address_county address_post_code>.freeze
 
--- a/app/services/additional_claimants_csv/model_builder.rb
+++ b/app/services/additional_claimants_csv/model_builder.rb
@@ -1,4 +1,5 @@
 class AdditionalClaimantsCsv::ModelBuilder
+
   ATTRIBUTES = %i<title first_name last_name date_of_birth address_building
     address_street address_locality address_county address_post_code>.freeze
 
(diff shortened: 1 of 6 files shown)


lib/errors/csv_dob_validation_error.rb
```diff
@@ -0,0 +1,2 @@
+class CsvDobValidationError < StandardError
```
leoapost: how about to move this class in the `app/services/additional_claimants_csv/validator.rb` and spare the new file and the extra path autoload?

zaparka: That sounds like a good idea :)

app/services/additional_claimants_csv/model_builder.rb
```diff
@@ -25,6 +26,13 @@ def attributes_from(row_data)
   end
 
   def sanitize(row_data)
+    check_csv_dob_format(row_data) if ATTRIBUTES.index(:date_of_birth)
```
leoapost: how about to use `.include?` instead of `.index` to make the statement easier to understand?

zaparka: done

app/services/additional_claimants_csv/validator.rb
```diff
@@ -1,3 +1,6 @@
+class CsvDobValidationError < StandardError
```
leoapost: maybe we could even move this into the `AdditionalClaimantsCsv::Validator` class:

```ruby
class AdditionalClaimantsCsv::Validator < AdditionalClaimantsCsv::BaseCsv
  class CsvDobValidationError < StandardError; end
  .
  .
  .
end
```

what do you think?

zaparka: That would limit that validation error to AdditionalClaimantsCsv class, which I want to avoid.

leoapost: ok, I see.

PR #855, opened 2017-05-26
Feature/gem updates

zaparka:
Ready to merge.


app/forms/address_attributes.rb
```diff
@@ -11,6 +11,8 @@ module AddressAttributes
     validates :address_post_code, post_code: true, length: { maximum: POSTCODE_LENGTH }
   end
 
+  # rubocop:disable MethodLength
```
zaparka: @leoapost I tried but there was a problem. I will revisit this when I have more time if that's ok with you?

Gemfile
```diff
@@ -47,6 +47,7 @@ group :development, :test do
   gem 'rspec-its', require: 'rspec/its'
   gem 'rspec-rails', '~> 3.0.0'
   gem 'rubocop'
```
zaparka: sure

PR #867, opened 2017-07-27
Removing payments from the user journey.

zaparka:


zaparka:
@leoapost yes it's hidden too.

leoapost:
Great. One more thing.. can we check the claim data that will be sent over to Jadu to ensure that there will be no fees or at least the fees will be zero?

PR #875, opened 2017-08-23
Refactoring Rubocop for spec/models

zaparka:


.rubocop.yml
```diff
@@ -21,8 +21,17 @@ AllCops:
     - 'lib/log_stuff.rb'
     - 'config/routes.rb'
     - 'Gemfile'
-    - 'spec/**/*'
```
leoapost: How about to keep this and use the `Include:` to include specific files/folders if needed?
```
Include:
  - 'spec/foo.rb'
  - 'spec/bar/**/*'
```

.rubocop.yml
```diff
@@ -173,6 +182,9 @@ Style/WordArray:
   Exclude:
```
leoapost: Can we also change the WordArray to enforce brackets?
```
Style/WordArray:
  EnforcedStyle: brackets
```

This will need to run `rubocop -a` to change the code.

.rubocop.yml
zaparka: @leoapost that's what I did but it didn't work.

.rubocop.yml
zaparka: ok

.rubocop.yml
leoapost: I see. Ok then.

PR #881, opened 2017-09-11
Feature/rst 449 - map web form to ET1 form

garytaylor:
Setup travis to use docker to allow our testing environment to be the same as theirs
Added cucumber test to prove the mapping from input forms through to PDF document
Added site prism gem for defining page objects - currently only used in cucumber tests
Added capybara screenshot to assist when things go wrong in the tests
uglifier gem now only needed in production environment, meaning nodejs is not required at runtime in development anymore
Now only using fog-aws gem instead of the entire fog collection for every other cloud provider out there.
database.yml can now get its values from the environment - useful for docker setups (defaults to old values so this will not break anyones dev setup)
rack live reload configured to use vendored version - prevents issues with docker setup
Carrierwave config now done in environment specific way to allow dev to be different to test
Carrierwave now uses file storage in test mode
Active job config now done in environment specific way
Active job config in test mode no longer uses inline - but test mode where it is up to the test code to perform any jobs at the time it wants to.  This is because inline mode is not real life - never would a job run in the same thread and in the same process, never mind at exactly the moment it was queued.
Admin routes now available to 172.x.x.x IP addresses for docker

Many tests now use a javascript driver for capybara (selenium - chrome) but there are still some that do not.
The reason for this is that there are issues in the application that prevent the test suite from passing when ran with a JS driver.  For the future, these need sorting and everything running with the JS driver, but for now, these will remain as a mixture.
selenium is configured for standard docker-compose setup by default but the url can be overridden using SELENIUM_URL for any docker haters out there - I didn't want to force you to have to use docker for testing - it does make life more predictable though, so would strongly encourage it


features/output_form.feature
```diff
@@ -0,0 +1,374 @@
+Feature: Output Form
```
AlexaVB: Is there a way to keep to 3 steps per scenario? i.e. Given, When, Then?

features/output_form.feature
```diff
@@ -0,0 +1,374 @@
...
+
+  @mock_jadu
+  Scenario: Scenario 1
```
AlexaVB: Can we give 'Scenario 1' a more descriptive title?

AlexaVB:
Discussed with Gary and Don, ticket has been raised for a refactor, to be completed in October.
Based on this, this PR can be merged.

PR #882, opened 2017-09-18
Fist part of removing Struct as a parent class.

zaparka:
Refactoring and removing some TODO comments. Mostly it's on inheritance from Struct where it's not needed.

### DIFF
--- a/app/presenters/confirmation_email_addresses_presenter.rb
+++ b/app/presenters/confirmation_email_addresses_presenter.rb
@@ -1,7 +1,11 @@
-# TODO: take a look at this rubocop warning
-# rubocop:disable Style/StructInheritance
-class ConfirmationEmailAddressesPresenter < Struct.new(:claim)
+class ConfirmationEmailAddressesPresenter
   delegate :primary_claimant, :representative, :secondary_claimants, to: :claim
+
+  attr_reader :claim
+
+  def initialize(claim)
+    @claim = claim
+  end
 
   def self.email_addresses_for(claim)
     new(claim).filter_email_addresses
@@ -26,4 +30,3 @@ class ConfirmationEmailAddressesPresenter < Struct.new(:claim)
     secondary_claimants.none? || representative.try(:email_address).blank?
   end
 end
-# rubocop:enable Style/StructInheritance
--- a/app/services/payment_gateway/request.rb
+++ b/app/services/payment_gateway/request.rb
@@ -1,5 +1,4 @@
 # Namespace for encapsulating Payment gateway related code
-# TODO: take a look at this rubocop warning
 # rubocop:disable Style/StructInheritance
 module PaymentGateway
   # Encapsulates ePDQ request related code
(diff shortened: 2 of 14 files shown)


app/presenters/confirmation_email_addresses_presenter.rb
```diff
@@ -1,8 +1,12 @@
-# TODO: take a look at this rubocop warning
-# rubocop:disable Style/StructInheritance
-class ConfirmationEmailAddressesPresenter < Struct.new(:claim)
+class ConfirmationEmailAddressesPresenter
   delegate :primary_claimant, :representative, :secondary_claimants, to: :claim
 
+  attr_reader :claim
+
+  def initialize(claim)
+    @claim = claim
```
garytaylor: I always see this as 'internal knowledge' of how attr_reader etc.. works - ie that it reads an instance variable and it writes to an instance variable.

I would normally do attr_writer :claim in the private section of the class, and call self.claim = claim in the initializer

The advantage is then that everything that changes stuff goes through a writer method - all in one place so easier to debug etc..

app/services/payment_gateway/request.rb
```diff
@@ -1,5 +1,4 @@
 # Namespace for encapsulating Payment gateway related code
-# TODO: take a look at this rubocop warning
 # rubocop:disable Style/StructInheritance
```
leoapost: Is this still needed?

spec/forms/additional_claimants_form_spec.rb
```diff
@@ -77,7 +77,7 @@
         additional_claimants_form.save
         claim.secondary_claimants.reload
 
-        attributes[:collection_attributes].each_with_index do |(_, attributes), index|
+        attributes[:collection_attributes].sort.each_with_index do |(_, attributes), index|
```
leoapost: Is this sort necessary? 

Same on line 99 👇

zaparka: @leoapost this is where the test was randomly failing. I think this is a fix for that.

app/services/payment_gateway/request.rb
zaparka: @leoapost I didn't remove the Struct from this one so yes it's still needed.

app/presenters/confirmation_email_addresses_presenter.rb
leoapost: @garytaylor can you please elaborate a bit more? I can't see how a writer method can make it easier to debug than a plain variable assignment.

zaparka: @garytaylor hmm that is an interesting way. I personaly never seen this approach. This implementation for me is a standard ruby way how to do this. @leoapost what do you think?

leoapost: I agree with @zaparka and I don't see any pros which would worth having private writer methods for instance variables.

I am happy to hear @garytaylor's approach regardless :)

garytaylor: @leoapost  - with a writer method - its 1 single place where the value gets written - not multiple.  So , say if you had a bug where something somewhere was changing this value.  With a writer method - simply override the default method, put a pry debug point in or whatever you use and you can see what is making the changes.  Without the writer method - do a search through the code, looking for anything that changes the instance var and put a pry debug point in there.

Its just the way ive always done it because I find it easier - but there isn't a standard ruby way of doing it - there are several !!

PR #883, opened 2017-09-29
WIP Rst 456 refund et fees

garytaylor:


app/forms/refunds/fees_form.rb
```diff
@@ -0,0 +1,133 @@
...
+    PAYMENT_METHODS = ['card', 'cheque', 'cash'].freeze
+    attribute :et_issue_fee,                            Float
+    attribute :et_issue_fee_payment_method,             String
+    attribute :et_issue_fee_payment_date,               Date
+    attribute :et_issue_fee_payment_date_unknown,       Boolean
+    attribute :et_hearing_fee,                          Float
+    attribute :et_hearing_fee_payment_method,           String
+    attribute :et_hearing_fee_payment_date,               Date
+    attribute :et_hearing_fee_payment_date_unknown,       Boolean
+    attribute :eat_issue_fee,                           Float
```
zaparka: @garytaylor what's an 'eat'?

app/forms/refunds/original_case_details_form.rb
```diff
@@ -0,0 +1,73 @@
+module Refunds
+  class OriginalCaseDetailsForm < Form
+    PAYMENT_METHODS = ['unknown', 'card', 'cheque', 'cash'].freeze
+    COUNTRY_OF_CLAIMS = ['england_and_wales', 'scotland'].freeze
```
zaparka: Sometimes there are different countries, but because we can't handle them we change the postcode to SW55 9QT. So you might look out for those.

config/routes.rb
```diff
@@ -30,6 +30,13 @@
       end
     end
 
+    resource :refund, only: [:create, :new], path: "/refund" do
+      RefundPagesManager.page_names.each do |page|
+        resource page.underscore, only: %i<show update>, controller: :refunds,
```
zaparka: I think we agreed not to use %i :)

db/migrate/20170831092724_create_refunds.rb
```diff
@@ -0,0 +1,74 @@
+class CreateRefunds < ActiveRecord::Migration
+  def change
+    create_table :refunds do |t|
```
zaparka: That's a big table.

app/forms/refunds/fees_form.rb
garytaylor: employment appeal tribunal

app/forms/refunds/original_case_details_form.rb
garytaylor: The country of claim is purely a question that we ask the user - meaning where was it issued - and there are only 2 places where they get issued from.  Its purely for admin purposed to allow them to contact the correct office about the claim thats  all so should be no issue

config/routes.rb
garytaylor: Well spotted - copied from original before the change I bet

db/migrate/20170831092724_create_refunds.rb
garytaylor: It is !!  Intentionally kept nice and flat for ease of exporting and so that associated addresses etc.. do not get mixed in with normal ET data which would have forced us to do things like polymorphic associations etc...

app/forms/refunds/original_case_details_form.rb
```diff
@@ -57,7 +57,7 @@ class OriginalCaseDetailsForm < Form
 
     def transfer_name
       claimant_name = [
-        resource.applicant_title.titleize,
+        resource.applicant_title&.titleize,
```
zaparka: What is the & for?

garytaylor: It allows for applicant_title being nil so titleize doesnt get called on nil

zaparka: ok so it's a shortcut for try.

app/models/session.rb
```diff
@@ -0,0 +1,19 @@
+class Session < ActiveRecord::Base
+
+  def to_h
+    data.dup
+  end
+
+  # rubocop:disable Style/MethodMissing
+  def method_missing(m, *args)
```
leoapost: I'm not a big fan of the `method_missing` unless it's being used in a framework or lib. I understand this implementation will give us setters/getters dynamically and have something like 
```
obj = Session.new
obj.name='Jane'
obj.post_code='E1'
obj.crazy_field='something crazy'
```
but you lose control of the Session's API, making hard to track misspelled calls like
```
obj.to_hh #data look up instead of warning me about a wrong method
=> nil
```
which can be worse if the class gets more methods.

Shouldn't something like the following be still simple enough and less error prone?
```
obj = Session.new
obj.data = {name: 'Jane', post_code: 'E1'} #set anything you like here
obj.add_data({crazy_field: 'something crazy'})
obj.data 
=> {:name => 'Jane', :post_code => 'E1', :crazy_field => 'something crazy'}
obj.data[:name]
=> "Jane"
```

db/migrate/20170831092724_create_refunds.rb
leoapost: Since these changes are permanent, I would suggest to not approach this as a quick solution. I think we should go through the new data we are going to introduce into the system and see if it makes sense to re-think their structure in the db.

@garytaylor is it possible to give us an idea of the new resources we are introducing with this feature?

leoapost:
# More suggestions

## JS files

Can we lint the js files? There are some style offences there like missing new lines between functions etc

---

How about to use double quotes in the jquery selectors to avoid escaping single quotes?
e.g.

```
$(".main-content.refunds_applicant [data-behavior='continue_button']")
```
instead of
```
$('.main-content.refunds_applicant [data-behavior=\'continue_button\']')
```

## Form files
Can we fix the styles like indentations?

## Disabling buttons styleguides

Based on the [gov.uk guides](http://govuk-elements.herokuapp.com/buttons/) we should disable buttons only when we have good reasons and the ET Refund form doesn't seem it needs it. If that's a requirement, could you please make sure the stake holders are aware of the guidelines? Maybe they missed it.

app/models/session.rb
leoapost: If we really like the `obj.name = 'Jane'` solution, then maybe it would worth consider using `OpenStruct`?

leoapost: Also, the class name is confusing. Maybe rename it to something like TempStorage? - I'm sure you can come up with a better name than me :)

todo.txt
```diff
@@ -0,0 +1,196 @@
+DONE Country of claim
```
leoapost: Probably we don't need this in the master branch, correct?

db/migrate/20171006134319_empty_refunds.rb
```diff
@@ -0,0 +1,13 @@
+class EmptyRefunds < ActiveRecord::Migration
```
leoapost: Do we need this migration?

docker-compose-test.yml
```diff
@@ -38,9 +38,11 @@ services:
       - npm_cache:/app/node_modules
     command: /bin/bash -c "bundle && npm install && rake db:create db:migrate assets:precompile cucumber"
   db:
-    image: postgres:9.5
+    image: postgres:9.3.5
   selenium:
     image: selenium/standalone-chrome-debug
+    volumes:
+      - /dev/shm:/dev/shm
```
leoapost: What's this volume for?

features/output_form.feature
```diff
@@ -63,30 +63,30 @@ Feature: Output Form
...
-      | title         | Mrs            |
-      | first_name    | Group5First    |
-      | last_name     | Group5Last     |
-      | date_of_birth | 21/10/1991     |
-      | building      | 112            |
-      | street        | Oxford Road    |
-      | locality      | London         |
-      | county        | Greater London |
-      | post_code     | SW1H 9JY       |
+#    And I choose to add more claimants
```
leoapost: Can we remove the commented out lines?

features/refund_error_recovery.feature
```diff
@@ -0,0 +1,25 @@
+Feature: Refund form with error recovery
+  Background:
+    Given I am "Anakin Skywalker"
+    # No town/city, county, email address in applicant page
```
leoapost: Can we remove the commented out lines?

features/refund_end_to_end_with_no_optionals.feature
```diff
@@ -0,0 +1,25 @@
+Feature: Refund form with no optional fields filled in
+  Background:
+    Given I am "Anakin Skywalker"
+    # No town/city, county, email address in applicant page
```
leoapost: Can we remove the commented out lines?

db/migrate/20170831092724_create_refunds.rb
garytaylor: The thing is, we are collecting information and not doing anything with it apart from exporting it, so converting into a relational structure for all the addresses, respondents etc... then converting it back again to export it seems completely pointless.

If we were doing anything with the data, then yes - we would do this a different way.

What is your motivation for wanting to do this differently ?

Gemfile
```diff
@@ -65,6 +65,10 @@ group :production do
   gem 'uglifier', '3.0.0'
 end
 
+group :production, :test do
```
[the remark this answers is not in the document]
garytaylor: Its not what you think - there are some unit tests that test that we do something with sentry (not my doing !! they were already in there)

Gemfile
```diff
@@ -65,6 +65,10 @@ group :production do
...
+  gem "sentry-raven", :git => "https://github.com/getsentry/raven-ruby.git"
```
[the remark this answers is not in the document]
garytaylor: Sorry, no idea - this is just what was in ET originally.

Procfile
```diff
@@ -0,0 +1,2 @@
+release: bundle exec rake db:migrate
```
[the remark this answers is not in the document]
garytaylor: I can't remove it until Dan has finished with it.  At the moment, it is useful to do quick reviews on any dev changes without me having to bother you to do a deploy all the time (my permissions in jenkins only allows me to see the build job)

app/forms/refunds/confirmation_form.rb
```diff
@@ -0,0 +1,11 @@
+module Refunds
+  class ConfirmationForm < Form
+    attribute :application_reference, String
+    attribute :submitted_at, Date
+
+    def initialize(resource, &block)
+      raise 'Session has no persisted refund' unless resource.respond_to?('_refund_id')
```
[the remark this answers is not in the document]
garytaylor: It shouldn't happen - it really is an exception, not a user error.  I simply put this in there to catch the error early if it did happen rather than await the confusion when it tries saving with no input data.
There is already a before action in the refunds controller that looks for missing input data (sessions table) and if its not present, the process is restarted with a flash message to the user

app/forms/refunds/fees_form.rb
```diff
@@ -0,0 +1,150 @@
...
+      self.eat_issue_fee_payment_date_without_pre_process = pre_process_partial_date(val, only: :day)
+    end
+    alias_method_chain :eat_issue_fee_payment_date=, :pre_process
+
+    def eat_hearing_fee_payment_date_with_pre_process=(val)
+      self.eat_hearing_fee_payment_date_without_pre_process = pre_process_partial_date(val, only: :day)
+    end
+    alias_method_chain :eat_hearing_fee_payment_date=, :pre_process
+
+    private
```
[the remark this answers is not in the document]
garytaylor: The headache is that there is a class method called 'dates' that does some weird stuff that looked ugly.  Hence, I could not use super here or anything - so had to use alias method chain which I hated, but the form stuff is not very nice code and does some over complex stuff.  I would prefer to spend the time upgrading to rails 5 and removing these awful form objects anyway - there are so many workarounds in the code because of it.

We could do a concern that does the alias method chain etc.. but the problem is, it would need to be included after the 'dates' class method call otherwise it would not work.

I tried really hard to keep this area clean, but the form object code was just causing issues all the time and this was the best I could come up with

app/models/refund.rb
```diff
@@ -0,0 +1,46 @@
+class Refund < ActiveRecord::Base
+  REFERENCE_START = 1000000
```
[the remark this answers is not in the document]
garytaylor: The reference numbers start at this value - this was requested functionality by the business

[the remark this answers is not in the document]
garytaylor: The reference number is given to the user and I guess they didnt want numbers like C1, C9 etc..

config/environments/heroku.rb
```diff
@@ -0,0 +1,55 @@
+Rails.application.configure do
```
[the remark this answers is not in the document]
garytaylor: Ill remove the heroku stuff when Im certain Dan has finished with it

config/environments/production.rb
```diff
@@ -1,3 +1,5 @@
+require 'raven'
```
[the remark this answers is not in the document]
garytaylor: I can't remember the reason, but we had to require it as it wasn't being required by bundler when the environments were being configured.  I can remove and see what breaks if you like ?

config/secrets.yml
```diff
@@ -13,6 +13,10 @@
 development:
   secret_key_base: 253145183f83ea3d0b9f39d188c123e4f71816fe802af7bb1cb44e60a3b45cfd4feb4900720b796f99a9a27b9f66b5f8bf3b39fbe26168364815ffa6dd30a05f
 
+heroku:
```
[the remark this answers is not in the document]
garytaylor: Again, will remove all heroku stuff once Dan has finished with it.

app/models/session.rb
garytaylor: We can't use open struct because it needs to be an AR object that persists to the DB. So I have just used a standard active record class that uses method missing so that you can get and set anything and persist to the database.  If you call the wrong method, you will get the wrong value - just like the underlying hash would.

We could make this so it us purely a 'Temporary Refund' and not general purpose and put the getters ad setters in - then rename the model and table name to suit.
But, would we not then be better off just having a table with columns again, but no DB level validation for null etc.. Even then, you  would look at the class and it would be empty just like normal AR classes without any validation etc...

garytaylor: I totally agree about the naming - when you first mentioned it, I thought 'Hmm, thats not a very good name really'.

Ill rename when we know if we want to change it to be more specific or not

todo.txt
garytaylor: This can now go yes

db/migrate/20171006134319_empty_refunds.rb
garytaylor: It was there to zap the data because I had a migration that changed the db level integrity checks (added null checking) - without this, the migration would fail.

Im not a big fan of changing migrations once they are in there, but maybe this one is an exception ?

docker-compose-test.yml
garytaylor: Shared memory for chrome - without this, chrome quickly runs out of shared memory and you get test failures - this is a known workaround

features/output_form.feature
garytaylor: Yes, well spotted - not sure if they need to be removed or re added - Ill check

features/refund_error_recovery.feature
garytaylor: We can, but I wanted them there to remind me what is behind 'Anakin Skywalker' - i.e. what is different about him

Gemfile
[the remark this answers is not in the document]
garytaylor: It doesnt send anything to sentry though as the environment variable isnt present

Gemfile
```diff
@@ -92,7 +95,7 @@ gem 'compass'
 gem 'virtus'
 gem 'zendesk_api'
 gem 'email_validator'
-gem "sentry-raven", :git => "https://github.com/getsentry/raven-ruby.git"
+gem 'logstasher', groups: [:production, :local]
```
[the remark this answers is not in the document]
garytaylor: Im not sure what the 'local' environment is actually used for - again, this was existing code not something Ive added so someone somewhere decided there was a reason for it - do you want me to remove ?

Procfile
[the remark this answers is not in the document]
garytaylor: I have now removed

app/forms/refunds/fees_form.rb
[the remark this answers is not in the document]
garytaylor: Ive now put the 'dates' class method call and the date transformation code into a concern to hide this complexity.

config/environments/heroku.rb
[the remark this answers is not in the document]
garytaylor: Removed

config/environments/production.rb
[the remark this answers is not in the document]
garytaylor: I removed it, then placed a breakpoint at the start of this file and the 'Raven' module was available - so committed this change

config/secrets.yml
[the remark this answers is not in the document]
garytaylor: Have now removed

Gemfile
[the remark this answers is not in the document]
leoapost: I only found the following three lines using Raven in our tests which are passing when I take the gem out of the test group. Are there any other tests that I may have missed?

```
spec/controllers/payments_controller_spec.rb
126:          expect(Raven).to receive(:capture_exception).with(message)

spec/forms/additional_claimants_form/additional_claimant_spec.rb
65:            expect(Raven).to receive(:extra_context).with(
75:            expect(Raven).to receive(:capture_exception)
```

Gemfile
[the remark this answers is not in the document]
leoapost: ok, let's leave it as is for now so to not get out of this task's scope. We can review the Gemfile later on.

app/models/refund.rb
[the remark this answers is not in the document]
leoapost: Great, thanks for the clarification.

config/secrets.yml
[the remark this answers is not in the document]
leoapost: Great, thanks :)

db/migrate/20171006134319_empty_refunds.rb
leoapost: Yes, I think it can be confusing for devs if we keep it. Also, since this does not affect the db schema then it's safe to remove.

Gemfile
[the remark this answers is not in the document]
garytaylor: Can you explain why those 3 tests would pass if the gem wasnt being loaded ?  I got const missing when I did it.
Originally, raven was in all groups so ive still improve it - if we wanted to get it out of test also, we would have to change these tests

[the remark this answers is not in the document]
garytaylor: Here is one of my failures

```
NameError: uninitialized constant Raven
./spec/controllers/payments_controller_spec.rb:126:in `block (5 levels) in <top (required)>'
-e:1:in `load'
-e:1:in `<main>'
```

app/models/session.rb
garytaylor: With the above method, we can add and remove columns from the refund schema etc.. without having to change this at all

db/migrate/20171006134319_empty_refunds.rb
garytaylor: have now removed it

Gemfile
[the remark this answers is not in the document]
leoapost: Sorry, my bad. I just noticed that I took it outside the groups (line 68) when I run the tests and bundler auto included it of course :)

app/forms/refunds/confirmation_form.rb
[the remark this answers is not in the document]
leoapost: OK, thanks for the info :)

app/models/session.rb
leoapost: Sorry, my example wasn't very clear. I meant to convert the data variable into an OpenStruct obj which would allow us to do `obj.data.name = 'Jane'` and `obj.data.name #=> Jane`.  If AR complains when saving it, then saving the `data.to_h` should do. Nevertheless, I would still prefer to use the simple and self explanatory `obj.data = {name: 'Jane'}`. 

Also, this method_missing will not only spend time and resources for the method lookups, but will confuse devs when calling wrong methods by returning nil values instead of error messages.

Now, all the above are just my thoughts around the method_missing and how to avoid issues, but if you think the possibilities of getting in trouble with the current implementation are low, then I am happy to keep it as is. :)

docker-compose-test.yml
leoapost: Nice, thanks for the info

app/models/session.rb
garytaylor: I know Ive replied to this on slack - but just for the record.  The form objects are written to use AR objects where the attributes are methods on the object and the save method is on the same object, also the valid? method etc...  I think it would get very hairy if we went down that route

PR #884, opened 2017-10-03
Loading activeadmin js only for activeadmin.

zaparka:
Making sure that we loading only what we need in application.js


leoapost:
@zaparka is this still relative?

zaparka:
@leoapost I totally forget about this one. @garytaylor do you think we still need this?

Stivaros:
Tidying up as this PR was inactive for a while. Sorry @zaparka 	:(

PR #887, opened 2017-10-13
Added link to paper refund form for when users have said their name has changed

garytaylor:
Added link to paper refund form for when users have said their name has changed

Nice simple one this time :-)

### DIFF
--- a/config/locales/refunds.yml
+++ b/config/locales/refunds.yml
@@ -160,7 +160,7 @@ en:
         For audit
       applicant_personal_information_usage_header: "Using your personal information:"
       applicant_personal_information_usage_sub_header: "Personal information which you supply to us may be used in a number of ways, for example:"
-      name_changed_cannot_continue: "If your name has changed since you submitted your original Employment Tribunal claim you will need to complete a paper form which can be downloaded here. You will be asked to provide evidence of your name change."
+      name_changed_cannot_continue_html: 'If your name has changed since you submitted your original Employment Tribunal claim you will need to complete a paper form which can be downloaded <a target="_blank" href="https://hmctsformfinder.justice.gov.uk/HMCTS/GetForm.do?original_id=2033">here</a>. You will be asked to provide evidence of your name change.'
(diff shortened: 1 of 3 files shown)


config/locales/refunds.yml
```diff
@@ -160,7 +160,7 @@ en:
         For audit
       applicant_personal_information_usage_header: "Using your personal information:"
       applicant_personal_information_usage_sub_header: "Personal information which you supply to us may be used in a number of ways, for example:"
-      name_changed_cannot_continue: "If your name has changed since you submitted your original Employment Tribunal claim you will need to complete a paper form which can be downloaded here. You will be asked to provide evidence of your name change."
+      name_changed_cannot_continue_html: 'If your name has changed since you submitted your original Employment Tribunal claim you will need to complete a paper form which can be downloaded <a target="_blank" href="https://hmctsformfinder.justice.gov.uk/HMCTS/GetForm.do?original_id=2033">here</a>. You will be asked to provide evidence of your name change.'
```
zaparka: Can you use some kind of helper method instead of having an HTML in translation file?

garytaylor: Html is in the yml file in a lot of places so we should put in a ticket for that one

leoapost: I get an `Error Page` when visiting https://hmctsformfinder.justice.gov.uk/HMCTS/GetForm.do?original_id=2033

garytaylor: Yes you will - dont worry about it - the content isnt done yet

leoapost: shouldn't that be a blocker then?

garytaylor: no - it will be in place by the time we put it live - we are just getting it ready.

garytaylor: Sometimes html in a translation file is good - what about if the link needed to change for a welsh version for example ?

leoapost: I've seen both approaches in our yaml files 
https://github.com/ministryofjustice/atet/blob/master/config/locales/en.yml#L460
https://github.com/ministryofjustice/atet/blob/master/config/locales/en.yml#L433
but I think I like a bit more the helper method way.

PR #888, opened 2017-10-17
Hotfix/minor updates refunds

garytaylor:
This PR is a bunch of back ported changes that were required to upgrade the application to rails 4.2.8 or above (up to rails 5.1).
They are all seen as valuable whether we choose to upgrade or not

### DIFF
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -1,6 +1,6 @@
 class FeedbackController < ApplicationController
   def create
-    resource.assign_attributes params[:feedback]
+    resource.assign_attributes params.require(:feedback)
     resource.save
 
     redirect_to feedback_path, flash: { info: t('.sent') }
--- a/app/controllers/payments_controller.rb
+++ b/app/controllers/payments_controller.rb
@@ -55,7 +55,7 @@ class PaymentsController < ApplicationController
       message = "Failed to recognize payment order: #{order}, status: #{status}"
       Raven.capture_exception(message)
       # Don't give anything away to attackers poking at the system
-      render nothing: true
+      head :ok
     end
   end
 
--- a/app/controllers/user_sessions_controller.rb
+++ b/app/controllers/user_sessions_controller.rb
@@ -23,7 +23,7 @@ class UserSessionsController < ApplicationController
   end
 
   def touch
-    render nothing: true
+    head :ok
   end
 
   def expired
--- a/app/controllers/user_sessions_controller.rb
+++ b/app/controllers/user_sessions_controller.rb
@@ -50,7 +50,13 @@ class UserSessionsController < ApplicationController
   end
 
   def user_session
-    @user_session ||= UserSession.new(params[:user_session])
+    @user_session ||= UserSession.new(user_session_params.to_h)
+  end
+
+  def user_session_params
+    params.require(:user_session).permit(:reference, :password, :email_address)
+  rescue ActionController::ParameterMissing
+    ActionController::Parameters.new
   end
 
   helper_method :user_session
--- a/app/controllers/user_sessions_controller.rb
+++ b/app/controllers/user_sessions_controller.rb
@@ -50,7 +50,7 @@ class UserSessionsController < ApplicationController
   end
 
   def user_session
-    @user_session ||= UserSession.new(user_session_params.to_h)
+    @user_session ||= UserSession.new(user_session_params.to_unsafe_hash)
   end
 
   def user_session_params
--- a/app/forms/collection_form.rb
+++ b/app/forms/collection_form.rb
@@ -22,6 +22,8 @@ class CollectionForm < Form
   end
 
   def collection_attributes=(attributes)
+    # Ensure we are working with a hash not action controller parameters
+    attributes = attributes.to_unsafe_hash if attributes.respond_to?(:to_unsafe_hash)
     attributes.each_with_index do |(_, resource_attributes), index|
       resource = relation[index] || relation.build
 
--- a/app/forms/collection_form.rb
+++ b/app/forms/collection_form.rb
@@ -53,7 +53,8 @@ class CollectionForm < Form
   end
 
   def valid?
-    run_callbacks(:validation) { super && collection.map(&:valid?).all? }
+    super # This will call validation callbacks etc... but ignore results
+    collection.all?(&:valid?)
   end
 
   def errors
--- a/app/forms/concerns/refunds/fees_form_date_transform.rb
+++ b/app/forms/concerns/refunds/fees_form_date_transform.rb
@@ -4,31 +4,29 @@ module Refunds
     included do
       dates :et_issue_fee_payment_date, :et_hearing_fee_payment_date, :eat_issue_fee_payment_date
       dates :eat_hearing_fee_payment_date, :et_reconsideration_fee_payment_date
-      alias_method_chain :et_issue_fee_payment_date=, :pre_process
-      alias_method_chain :et_hearing_fee_payment_date=, :pre_process
-      alias_method_chain :et_reconsideration_fee_payment_date=, :pre_process
-      alias_method_chain :eat_issue_fee_payment_date=, :pre_process
-      alias_method_chain :eat_hearing_fee_payment_date=, :pre_process
+      prepend PreProcess
     end
 
-    def et_issue_fee_payment_date_with_pre_process=(val)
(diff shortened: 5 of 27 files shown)


app/controllers/feedback_controller.rb
```diff
@@ -1,6 +1,6 @@
 class FeedbackController < ApplicationController
   def create
-    resource.assign_attributes params[:feedback]
+    resource.assign_attributes params.require(:feedback)
```
garytaylor: Rails is getting stricter with regards to parameters from the controller, so we will need to use strong parameters like this in the future.  It also adds to the warm feeling of knowing what we are allowing through etc..

app/controllers/payments_controller.rb
```diff
@@ -55,7 +55,7 @@ def validate_request
       message = "Failed to recognize payment order: #{order}, status: #{status}"
       Raven.capture_exception(message)
       # Don't give anything away to attackers poking at the system
-      render nothing: true
```
garytaylor: render nothing: true  or render :nothing is going in rails 5.1 - so now we are ready

app/controllers/user_sessions_controller.rb
```diff
@@ -50,7 +50,13 @@ def deliver_access_details
   end
 
   def user_session
-    @user_session ||= UserSession.new(params[:user_session])
+    @user_session ||= UserSession.new(user_session_params.to_unsafe_hash)
```
garytaylor: Not as bad is it sounds - it is converting a set of parameters that have been permitted - to a hash.  The word unsafe is misleading

app/controllers/user_sessions_controller.rb
```diff
@@ -50,7 +50,13 @@ def deliver_access_details
...
+  end
+
+  def user_session_params
+    params.require(:user_session).permit(:reference, :password, :email_address)
```
garytaylor: user_session is optional - if it is missing we simply want an empty set of params.  This is the original design, just had to stick with it

app/forms/collection_form.rb
```diff
@@ -53,7 +55,8 @@ def save
   end
 
   def valid?
-    run_callbacks(:validation) { super && collection.map(&:valid?).all? }
```
garytaylor: The valid? method was previously flawed in that it was considering its own validity in the equation - however, in rails 5 - its own validity comes into question because the errors for each item in the collection are stored in the errors object even though they are empty, but the valid check simply checks for zero length.

The previous version also called the callbacks twice

This change simply checks for validity on all of the sub forms

app/forms/concerns/refunds/fees_form_date_transform.rb
```diff
@@ -4,31 +4,29 @@ module FeesFormDateTransform
     included do
       dates :et_issue_fee_payment_date, :et_hearing_fee_payment_date, :eat_issue_fee_payment_date
       dates :eat_hearing_fee_payment_date, :et_reconsideration_fee_payment_date
-      alias_method_chain :et_issue_fee_payment_date=, :pre_process
```
garytaylor: Alias method chain is old hat now - replaced with Module#prepend

app/forms/form/date_parsing.rb
```diff
@@ -14,7 +14,7 @@ def dates(*dates)
 
             define_method("#{date}=") do |obj|
               begin
-                super coerce_object(obj)
+                super coerce_object(normalize_date_params(obj))
```
garytaylor: Just some refactoring as guided by rubocop

app/forms/refunds/applicant_form.rb
```diff
@@ -44,7 +44,7 @@ class ApplicantForm < Form
                               length: { maximum: EMAIL_ADDRESS_LENGTH }
     validates :applicant_address_telephone_number, presence: true
     validates :applicant_date_of_birth, presence: true
-    validates :has_name_changed, nil: true
```
garytaylor: In rails 5, the Yes / No radio button group sends an empty string in the params, previously nothing was sent so the value was nil when unanswered.  
This validator is now nil OR empty string, both meaning the same thing - 'Unanswered question'

app/inputs/date_with_unknown_input.rb
```diff
@@ -65,9 +65,10 @@ def build_checkbox_field(buffer)
 
   def value
     value = object.send attribute_name
-
     if value.is_a? Hash
       OpenStruct.new value
+    elsif value.is_a? ActionController::Parameters
```
garytaylor: Can now cope with action controller parameters (which no longer extend a hash)

app/validators/additional_claimants_csv_validator.rb
```diff
@@ -18,7 +18,7 @@ def result(record)
   end
 
   def add_errors(record, attribute)
-    record.errors[:erroneous_line_number] = @result.error_line.to_s
```
garytaylor: Deprecated API

spec/controllers/feedback_controller_spec.rb
```diff
@@ -11,7 +11,7 @@
     before { allow(FeedbackForm).to receive(:new).and_return form }
 
     it 'saves the form' do
-      expect(form).to receive(:assign_attributes).with(params).ordered
+      expect(form).to receive(:assign_attributes).with(an_object_having_attributes(to_unsafe_hash: params.stringify_keys))
```
garytaylor: Expected action controller parameters, not a hash

spec/features/multiple_respondents_spec.rb
```diff
@@ -70,7 +70,7 @@
       end
 
       click_button 'Save and continue'
-
+      claim.reload
```
garytaylor: We must not assume claim is the same object as the web server modified

spec/forms/additional_claimants_form_spec.rb
```diff
@@ -32,7 +32,7 @@
       additional_claimants_form.assign_attributes attributes
     end
 
-    let(:collection) { [Claimant.new, Claimant.new] }
+    let(:collection) { [Claimant.new(claim_id: claim.id), Claimant.new(claim_id: claim.id)] }
```
garytaylor: The claimant belongs to the claim - so not realistic unless this is populated

spec/forms/additional_claimants_form_spec.rb
```diff
@@ -44,7 +44,7 @@
 
     it 'decorates AdditionalClaimants as Claimants' do
       additional_claimants_form.collection.each_with_index do |_c, i|
-        expect(additional_claimants_form.collection[i].target).to eq collection[i]
+        expect(additional_claimants_form.collection[i].target.as_json).to eq collection[i].as_json
```
garytaylor: For some reason, in rails 5, comparing AR objects as equal did not work even if the attributes were the same, so now comparing the results of to_json so the test is simpler

spec/forms/respondent_form_spec.rb
```diff
@@ -154,7 +154,7 @@
     before { allow(respondent_form.target).to receive :enqueue_fee_group_reference_request }
 
     it 'addresses reloaded on save' do
-      expect(respondent_form.target.addresses).to receive(:reload)
+      expect_any_instance_of(respondent_form.target.addresses.class).to receive(:reload)
```
garytaylor: We cannot assume that respondent_form.target.addresses is going to be the same object all the time - in rails 5 it isn't - we get a new proxy object each time.
We should not really be mocking interfaces that are not ours, but the old test did, so i have simply made this test work - not ideal using expect_any_instance_of, but without re designing the interfaces, this was the best I could achieve

spec/forms/refunds/applicant_form_spec.rb
```diff
@@ -204,10 +210,10 @@ module Refunds
         expect(applicant_form.applicant_date_of_birth).to be value
       end
 
-      it 'stores the provided ActionController::Parameters if the date is invalid' do
-        value = ActionController::Parameters.new('day' => '32', 'month' => '15', "year" => '1985')
+      it 'stores the provided ActionController::Parameters as a hash if the date is invalid' do
+        value = ActionController::Parameters.new('day' => '32', 'month' => '15', "year" => '1985').freeze
```
zaparka: Why are you freezing this?

garytaylor: To be sure nothing changes it - I was concerned that the underlying code might modify the existing params rather than convert to a new set or a hash

PR #890, opened 2017-10-24
Feature/rst542 nokogiri update

garytaylor:
Nokogiri required an upgrade, however, our application was not requesting it as a dependency - rails itself was - so it did not seem right to just add it to our gemfile and force rails to have a later version than what it was tested with at 4.2.0.  I noticed when I did an upgrade to 4.2.1 recently that nokogiri was upgraded, so it made sense to keep this upgrade.
We are only upgrading from 4.2.0 - a safe minor version increment - all tests pass and the system works as before.


config/environments/development.rb
```diff
@@ -32,6 +32,10 @@
   # number of complex assets.
   config.assets.debug = true
 
+  # Asset digests allow you to set far-future HTTP expiration dates on all assets,
+  # yet still be able to expire them through the digest params.
+  config.assets.digest = true
```
leoapost: Do we need the digest in dev mode?

Gemfile
```diff
@@ -10,7 +10,7 @@ gem 'jbuilder', '~> 2.0'
 gem 'jquery-rails', '4.3.1'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '4.2.0'
+gem 'rails', '4.2.1'
```
leoapost: Do you think we can use the latest rails v4.2.x to benefit from the bug fixes and security patches?

config/environments/development.rb
garytaylor: It was just what came from the newer version of rails.  It can still be handy IMO depending on browser.

Gemfile
garytaylor: I have ran tests against all versions up to 4.2.10 (latest I believe of 4.2 branch) and all is fine.  Ive even got 5 working but thats probably going too far

PR #900, opened 2017-11-14
Feature/rst 623 csv export

zaparka:


app/models/refund.rb
```diff
@@ -43,4 +43,40 @@ def eat_issue_fee_present?
   def eat_hearing_fee_present?
     eat_hearing_fee.present? && eat_hearing_fee.positive?
   end
+
+  class << self
+    def to_csv(ids)
```
garytaylor: Im not sure I follow this.  Where do ids come from ?  If there is an id for every record in the database, what happens when there are 50,000 of them ? 
The csv is being built in memory rather than streaming to a file directly - again, if we are dealing with 50,000 records, I would be a bit worried about memory usage.  I guess we would need to estimate the size of an average row ?

spec/models/refund_spec.rb
```diff
@@ -0,0 +1,199 @@
+require 'rails_helper'
```
garytaylor: Thinking about it, Im not sure exporting belongs in a model ?  Isn't that what we use services for in our projects ?

app/models/refund.rb
zaparka: @garytaylor the IDs are from the view. You can't select 50k records. But I guess that you can call the method from another one. But I'm using ```find_each``` method which is iterating in batches. So I would not worry about the memory usage because that's what find_each is for.

spec/models/refund_spec.rb
zaparka: Hmm, interesting thought. I'm not sure. I'm calling to_csv which has the same principle as to_h or to_s. Then I'm using presenter, so we don't have that code in the model but in a presenter. But I'm open to suggestions. @leoapost what do you think?

app/models/refund.rb
garytaylor: Does this mean the user can only export a page at a time ?  If so, is that acceptable ?

spec/models/refund_spec.rb
garytaylor: I think whether right or wrong (although its never wrong if it works lol) - I guess as long as its consistent with other areas of the system - i.e. if other areas use services then we should use services

app/models/refund.rb
zaparka: @garytaylor yes that's what it means. There is no AC that would say that it has to be exported all at once by the user. My thinking was to show it to Cathie so I can get some feedback.

zaparka: Ok. let's use service.

spec/models/refund_spec.rb
zaparka: Ok. let's use service.

app/models/refund.rb
garytaylor: ok - I would have expected it to export all records, but if that wasnt specified then maybe not - like you say, check with Cathie

leoapost:
@zaparka I take it this PR is not needed anymore. Can we close it?

zaparka:
the functionality is not needed anymore.

PR #909, opened 2018-01-30
Adding info notes to ACAS form.

zaparka:
To reduce incorrectly checked in "I don't have an Acas number" we are adding more information to the form.

### DIFF
--- a/app/assets/javascripts/modules/moj.reveal-hints.js
+++ b/app/assets/javascripts/modules/moj.reveal-hints.js
@@ -12,3 +12,19 @@
     }
   });
 })();
+
+(function() {
+  $('.panel-indent input[type=radio]').each(function(i, field){
+    var container = $(field);
+    if(container.length){
+      container.on('change', function(){
+        var hint = $(this).parents('.panel-indent').children('.reveal-acas-hint');
+        if(this.value == 'interim_relief' && this.checked) {
+          hint.show();
+        } else {
+          hint.hide();
+        }
+      });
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/modules/moj.reveal-hints.js
```diff
@@ -11,4 +11,20 @@
...
     }
   });
+})();
+
+(function() {
+  $('.panel-indent input[type=radio]').each(function(i, field){
+    var container = $(field);
+    if(container.length){
+      container.on('change', function(){
+        var hint = $(this).parents('.panel-indent').children('.reveal-acas-hint');
```
garytaylor: is this right ?  I thought this was an mot module which is global - but it has something about 'acas hints' in it ?

zaparka: @garytaylor this is a module that reveals and hide all hints, but I can't use it. The view is using collection so I can't dynamically assign data attribute based on the checkbox value. So I thought that this is the best place to put it. I can use a new file, but It makes sense to me to keep it together with similar functionality. But if you would insist on putting into another file, I don't mind.

garytaylor: No not at all - just wanted to check it wasn't a mistake thats all - if you're happy with it in there so am I .

zaparka: cool thx

PR #912, opened 2018-03-21
The db host is now configurable via the environment

garytaylor:
The db host is now configurable via the environment for use with docker setups using the 'local' environment.


AlexaVB:
Looks good @garytaylor

PR #916, opened 2018-05-17
Updating refund details page

zaparka:
Removing reference of building society because Liberata is not using it anyway.
DO NOT MERGE


zaparka:
Renaming the branch

PR #923, opened 2018-06-07
Feature/rst 1113 diversity form

zaparka:
This is implementation of Diversity questioanoire.


garytaylor:
Hi @zaparka - The diversity form is supposed to send its data to the api for storage instead of / as well as storing it itself.
Remember I mentioned we needed to have a chat about the api - as its not defined yet.
Or, are you doing this as a separate piece of work ?

zaparka:
@garytaylor this branch is not for the API bit. it's just for getting and storing the data to DB. There will be different PR for that API bit.

PR #927, opened 2018-07-16
Feature/rst 1115 json api

zaparka:
This is API part of diversity work.


zaparka:
This is already merged to master

PR #933, opened 2018-09-14
Feature/rst 1346 translation

zaparka:
Welsh Translation for Claim and Diversity.


app/controllers/errors_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def not_found
+    render(status: 404)
+  end
+
+  def unprocessable
+    render(status: 422)
+  end
+
+  def internal_server_error
```
garytaylor: Just spotted this - is this not a bit dangerous ?  If you have a 500 error because of something inside the app that happens to generate a 500 error even when this view is rendered - we would end up in a loop..  I think this is why generally these error pages are always static pages

zaparka: @garytaylor you wouldn't get in the loop. I tried. You would get just another error page but without any helpful information or formatting. But I know what you mean. I using dynamic error page so we use the translation version for them. The other way would be to have a static page for every error in 2 languages. I'm happy to talk about it,

garytaylor: Im probably not the best person to talk to about it - maybe one of the devops guys - Isher or Ben might have an opinion ?

config/locales/cy/claim_review.cy.yml
```diff
@@ -0,0 +1,137 @@
...
+
+      your_fee:
+        seeking_remission: "Help i dalu ffioedd "
+        fee: Eich ffi
+
+    email_addresses:
+      submit_warning_header: Cyflwyno eich hawliad
+      submit_warning: "Hwn yw'r cyfle olaf i chi newid manylion eich hawliad ar-lein. Gwiriwch fod yr holl wybodaeth uchod yn gywir. "
+      confirmation: Cadarnhad drwy e-bost
+      submit_with_remission: "Cyflwyno'r hawliad "
```
garytaylor: Have you noticed the space at the end ?  Ive spotted this in a few places

zaparka: @garytaylor  well spotted. Removing now.

PR #944, opened 2018-10-23
RST-1469: Bootsnap in Production Only

Stivaros:
Bootsnap is used to cache operations and increase boot time. This can make debugging initialisers a pain in the backside, so this PR moves bootsnap to prod only.


Stivaros:
I will merge this after Welsh is released. Prevents another rebase in that branch.

PR #945, opened 2018-10-25
Fixing failing tests after locale moved to url

zaparka:
There is a requirement for a locale to be part of the URL. This is the PR that does just that. The url without locale in it will still work.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,75 +6,83 @@ Rails.application.routes.draw do
   match "/500", :to => "errors#internal_server_error", :via => :all
   match "/503", :to => "errors#service_unavailable", :via => :all
 
-  scope :apply do
-    resource :guide,              only: :show
-    resource :terms,              only: :show
-    resource :cookies,            only: :show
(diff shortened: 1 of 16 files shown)


config/routes.rb
```diff
@@ -6,75 +6,83 @@
...
-    resource :claim_review,       only: %i<show update>, path: :review
-    resource :pdf,                only: :show
-    resource :claim_confirmation, only: :show, path: :confirmation
-
-    resource :claim, only: :create, path: "/" do
-      resource :payment, only: %i<show update>, path: :pay do
-        member do
-          %i<success decline>.each do |result|
-            get result, to: "payments##{result}", as: result
+  scope "(:locale)", locale: /en|cy/ do
```
garytaylor: Can this default to en if its not present ?  I did this in the calculator but cannot remember how !
If we dont do this, if someone is in the middle of a session when we deploy it, they will get a 500 error on the following page.
Also, if anyone has things bookmarked, they will break
Any google results that are cached would also break

PR #949, opened 2018-11-08
Ensured all multi choice checkbox groups have had their readonly attr…

garytaylor:
Ensured all multi choice checkbox groups have had their readonly attribute removed

Simple form can set the readonly attribute based on the values of the input data - this ensures that none of the radio button groups have the readonly attribute set

### DIFF
--- a/app/views/diversities/_claim_type.html.slim
+++ b/app/views/diversities/_claim_type.html.slim
@@ -1,7 +1,7 @@
 fieldset
   = f.input :claim_type,
     as: :radio_buttons,
-    include_hidden: false,
+    readonly: nil,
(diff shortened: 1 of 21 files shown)


app/views/diversities/_claim_type.html.slim
```diff
@@ -1,7 +1,7 @@
 fieldset
   = f.input :claim_type,
     as: :radio_buttons,
-    include_hidden: false,
+    readonly: nil,
```
Stivaros: Have you intentionally overwritten `include_hidden: false` on this one?

garytaylor: it was a duplicate

PR #1152, opened 2019-11-25
Rst 2229 validation

Fred83200:
I added validation on the employment date.
So the end date can't be before the start date.

```
[ ] Yes
[ X] No
```


app/forms/employment_form.rb
```diff
@@ -94,4 +95,12 @@ def reset_new_job!
   def target
     resource.employment || resource.build_employment
   end
+
+  def end_date_after_start_date?
+    return if end_date.blank? || start_date.blank?
+
+    if end_date < start_date
+      errors.add(:end_date)
```
garytaylor: What happened to the second parameter ? the one specifying what the error is (:end_date_before_start)

config/locales/cy/cy.yml
```diff
@@ -378,6 +378,7 @@ cy:
         net_pay: "Hwn yw eich tâl net neu dâl terfynol, ar ôl treth a didyniadau eraill. Gallwch ddod o hyd iddo ar eich slip cyflog. Dylech gynnwys unrhyw daliadau goramser, comisiwn a bonws"
         benefit_details: Manylion budd(ion)
         end_date: "Er enghraifft, 22 04 2014 (os nad ydych yn gwybod yr union ddyddiad, rhowch eich amcangyfrif gorau)"
+        end_date_before_start: Rhaid nodi dyddiad sydd ar ôl dyddiad cychwyn y gyflogaeth
```
garytaylor: Is this a mistake ?  The correct place is line 860 which you have - I doubt this one is used

config/locales/en/en.yml
```diff
@@ -319,6 +320,7 @@ en:
         net_pay: This is your net or take-home pay, after tax and other deductions. You can find it on your payslip. Include overtime, commission and bonuses
         benefit_details: Details of benefit(s)
         end_date: For example, 22 04 2014 (if you don’t know the exact date then put your best estimate)
+        end_date_before_start: must be after the employment start date
```
garytaylor: Again, I think this is a mistake - should only be required in one place

PR #1153, opened 2019-11-25
Rst1124-Misleading-Email

Fred83200:
The changes has been made for the email field to appear only when email has been selected.

```
[ ] Yes
[ X] No
```


app/views/claims/_claimant.html.slim
```diff
@@ -35,14 +35,20 @@ fieldset
   = f.input :address_country, collection: ClaimantForm::COUNTRIES, include_blank: false
   = f.input :address_telephone_number, as: :tel, input_html: { class: 'medium-input' }
   = f.input :mobile_number, as: :tel, input_html: { class: 'medium-input' }
-  = f.input :email_address
 
-  fieldset
```
garytaylor: You have removed the fieldset from the contact_preference question.  Not sure if this is intentional ? I think it should have the fieldset around it

config/initializers/simple_form.rb
```diff
@@ -47,6 +47,18 @@
     b.use :input
   end
 
+  # vertical input for radio buttons and check boxes
```
garytaylor: Be really careful with this one.  As you are changing global configuration, this has the potential to change all forms and you might not notice.  
Can you validate that it hasnt please ? (literally go through the browser  html and compare before and after around the form elements) or if there is another solution to the problem you were trying to solve that doesnt involve this global change, that would be better.

app/views/claims/_claimant.html.slim
```diff
@@ -35,14 +35,22 @@ fieldset
   = f.input :address_country, collection: ClaimantForm::COUNTRIES, include_blank: false
   = f.input :address_telephone_number, as: :tel, input_html: { class: 'medium-input' }
   = f.input :mobile_number, as: :tel, input_html: { class: 'medium-input' }
-  = f.input :email_address
+
+  = f.input :contact_preference,
```
garytaylor: The fieldset is definitely missing from around the contact preferences.  Github is not showing the change very well, but if you check in the develop branch yourself, you will see that the contact_preference field does have a fieldset around it ?

PR #1355, opened 2021-02-24
Rst 2803 - Diversity

Fred83200:
Upgraded the diversity questionnaire with the GDS component.

```
[ ] Yes
[X] No
```


garytaylor (approved):
Ive approved as nothing major that we cant tidy up later - but have a look at the comments and if you want to fix now then please do but only if quick - if it is gonna take longer than another day then lets leave it as is.

PR #1375, opened 2021-03-19
RST-3153 ccd acas

Fred83200:
Added the ccd_acas_exemption_reason validation with spec and validator.

```
[ ] Yes
[ X] No
```

### DIFF
--- a/app/forms/respondent_form.rb
+++ b/app/forms/respondent_form.rb
@@ -57,8 +57,8 @@ class RespondentForm < Form
 
   validates :no_acas_number_reason,
     inclusion: { in: NO_ACAS_REASON, allow_blank: true },
-    presence: { if: -> { no_acas_number? } },
-            ccd_acas_exemption_reason: true
+            ccd_acas_exemption_reason: { if: :no_acas_number? },
+    presence: { if: -> { no_acas_number? } }
 
   validates :acas_early_conciliation_certificate_number,
     presence: { unless: -> { no_acas_number? } },
--- a/app/forms/respondent_form.rb
+++ b/app/forms/respondent_form.rb
@@ -57,7 +57,8 @@ class RespondentForm < Form
 
   validates :no_acas_number_reason,
     inclusion: { in: NO_ACAS_REASON, allow_blank: true },
-    presence: { if: -> { no_acas_number? } }
+    presence: { if: -> { no_acas_number? } },
+            ccd_acas_exemption_reason: true
 
   validates :acas_early_conciliation_certificate_number,
     presence: { unless: -> { no_acas_number? } },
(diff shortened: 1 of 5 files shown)


app/forms/respondent_form.rb
```diff
@@ -57,7 +57,8 @@ class RespondentForm < Form
 
   validates :no_acas_number_reason,
     inclusion: { in: NO_ACAS_REASON, allow_blank: true },
-    presence: { if: -> { no_acas_number? } }
+    presence: { if: -> { no_acas_number? } },
+            ccd_acas_exemption_reason: true
```
garytaylor, line 61: have you tried this ?  If you specify an acas number I think it will still try to validate the ccd acas exemption reason

config/locales/cy/cy.yml
```diff
@@ -974,6 +974,7 @@ cy:
             no_acas_number_reason:
               blank: "Dewiswch reswm o'r rhestr."
               inclusion: ""
+            invalid_cdd_acas_exemption_reason: 'activemodel.errors.models.respondent.attributes.no_acas_number_reason.blank'
```
garytaylor, line 977: Have you looked at the error message that is output from this ?  I think you need a colon in front of the 'activemodel.....' - the colon says that it is a reference to another entry rather than the value being 'activemodel.errors.models.respondent.attributes.no_acas_number_reason.blank'

spec/validators/ccd_acas_exemption_reason_validator_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        model = model_class.new(no_acas_number_reason: value.downcase)
+
+        model.valid?
+
+        expect(model.errors).to be_empty, "expected #{value} to be valid but it wasn't"
+      end
+    end
+  end
+
+  it 'is not valid for nil' do
```
garytaylor, line 47: I believe that this value can be nil - it has to be nil when an acas number is given - as there will be no reason to give

spec/validators/ccd_acas_exemption_reason_validator_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    model = model_class.new(no_acas_number_reason: nil)
+
+    model.valid?
+
+    expect(model.errors.details[:no_acas_number_reason]).to include a_hash_including(error: :invalid_cdd_acas_exemption_reason)
+  end
+
+  it 'is not valid for an empty string' do
```
garytaylor, line 55: this is still valid though - whilst the value can be nil - it cant be an empty string

garytaylor (changes requested):
Please see comments for details - unsure if you have tried this in the browser with all combinations of having an acas number, not having one - and all the various reasons

garytaylor (approved):

PR #1377, opened 2021-03-24
RST-3153 ccd address fields

Fred83200:
Added test and validator for the address field 

```
[ ] Yes
[X] No
```

### DIFF
--- a/config/locales/cy/cy.yml
+++ b/config/locales/cy/cy.yml
@@ -985,7 +985,6 @@ cy:
               blank: Mae'n rhaid i chi roi gwybodaeth ar gyfer y maes hwn
             address_building:
               blank: Rhowch enw neu rif adeilad cyfeiriad yr atebydd
-            blank_ccd_address: :'activemodel.errors.models.respondent.attributes.address_building.blank'
             address_street:
               blank: Rhowch stryd cyfeiriad yr atebydd
             address_locality:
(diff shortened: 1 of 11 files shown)


spec/validators/ccd_address_validator_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        model = model_class.new(example_address_street: value.downcase)
+
+        model.valid?
+
+        expect(model.errors).to be_empty, "expected #{value} to be valid but it wasn't"
+      end
+    end
+  end
+
+  it 'is not valid for nil' do
```
garytaylor, line 47: The scenario name says it should not be valid but the scenario itself says its testing for no errors

spec/validators/ccd_address_validator_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    model = model_class.new(example_address_street: nil)
+
+    model.valid?
+
+    expect(model.errors.details[:example_address_street]).to be_empty
+  end
+
+  it 'is not valid for an empty string' do
```
garytaylor, line 55: Same as above comment

garytaylor (changes requested):
Just a couple of minor issues

config/locales/cy/cy.yml
```diff
@@ -984,6 +985,7 @@ cy:
               blank: Mae'n rhaid i chi roi gwybodaeth ar gyfer y maes hwn
             address_building:
               blank: Rhowch enw neu rif adeilad cyfeiriad yr atebydd
+            blank_ccd_address: :'activemodel.errors.models.respondent.attributes.address_building.blank'
```
garytaylor, line 988: I cannot see anywhere this is being used ?  The correct message is invalid_ccd_address ?

garytaylor (approved):

PR #1386, opened 2021-04-06
Rst 3231

Fred83200:
Updated the timeout session views for the GDS component upgrade

```
[ ] Yes
[X] No
```

### DIFF
--- a/app/views/timeout_sessions/edit.html.slim
+++ b/app/views/timeout_sessions/edit.html.slim
@@ -1,5 +1,5 @@
 p= t('claims.show.page_number', current_page: current_page, total_pages: total_pages)
-h1.govuk-heading-xl= t("claims.registration.header")
+h1.govuk-heading-xl= t(".header")
 
 = form_for resource, as: user_session, url: user_session_path, method: :post, builder: EtGdsDesignSystem.form_builder_class do |f|
   = hidden_field_tag :return_to_review, params[:return_to_review]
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/active_admin.js
```diff
@@ -0,0 +1 @@
+//= require active_admin/base
```
garytaylor, line 1: What is this for ?  we are now using webpacker instead of sprockets so shouldn't be using this

app/views/timeout_sessions/edit.html.slim
```diff
@@ -1,35 +1,25 @@
...
+p= t('claims.show.page_number', current_page: current_page, total_pages: total_pages)
+h1.govuk-heading-xl= t(".header")
+
+= form_for resource, as: user_session, url: user_session_path, method: :post, builder: EtGdsDesignSystem.form_builder_class do |f|
+  = hidden_field_tag :return_to_review, params[:return_to_review]
+  = f.govuk_error_summary(t 'shared.error_notification.default_message')
+  = f.govuk_fieldset label: {text: t('timeout_sessions.registration.application_number')} do
+    span.number= claim.reference
+
+    = f.govuk_fieldset label: {text: false} do
```
garytaylor, line 10: whats the use of this blank fieldset ?  In general, if you find yourself telling gds to not label something - then accessibility side of things will struggle - how does a screen reader read it if its not labelled

PR #1413, opened 2021-06-15
Rst 3230

Fred83200:
GDS upgrade for feedback page

```
[ ] Yes
[ X] No
```


app/views/feedback/new.html.slim
```diff
@@ -1,21 +1,15 @@
...
       .form-fields
         = f.invisible_captcha :subtitle
-        = f.input :comments,
-          as: :text,
-          input_html: { rows: 5 }
-        = f.input :suggestions,
-          as: :text,
-          input_html: { rows: 5 }
-        = f.input :email_address
+        = f.text_area :comments, label: { text: "#{t('claims.feedback.feedback_comments')}"}, max_chars: 2500, rows: 5
```
garytaylor, line 9: You dont need label: { text: <blablabla> }

You can just do label: <blablabla> when using the methods that are not prefixed with govuk_

garytaylor (approved):

PR #1467, opened 2022-01-12
RST-2040 Radio Buttons

Fred83200:
Changes to the optional Radio Buttons

```
[ ] Yes
[ ] No
```


Gemfile.lock
```diff
@@ -603,4 +603,4 @@ DEPENDENCIES
   webrick (~> 1.4, >= 1.4.2)
 
 BUNDLED WITH
-   1.17.3
+   2.1.4
```
garytaylor, line 606: There is no need for this change.  We will need to upgrade bundler in the docker images to support this

config/locales/cy/cy.yml
```diff
@@ -71,4 +71,42 @@ cy:
       password:
         label: Crëwch eich gair cofiadwy
         hint: "Os ydych wedi nodi eich cyfeiriad e-bost uchod, byddwch yn gallu ailosod eich gair cofiadwy os byddwch yn ei anghofio"
+  activemodel:
```
garytaylor, line 74: I think you have put this in an inconsistent file when compared with the English translation file for active model errors (config/locales/en/et1/active_model.en.yml)

db/migrate/20220113123019_migrate_other_known_claimants_in_claims.rb
```diff
@@ -0,0 +1,15 @@
+class MigrateOtherKnownClaimantsInClaims < ActiveRecord::Migration[6.1]
+  class Claims < ActiveRecord::Base
+    self.table_name = :claims
+  end
+
+  def up
+    Claim.all.each do |claim|
+      claim.update_columns other_known_claimants: claim.other_known_claimants.present?
```
garytaylor, line 8: This looks incorrect - the original value of this was calculated in this method


```ruby

  def other_known_claimants
     @other_known_claimants ||= other_known_claimant_names?
   end

```

db/migrate/20220114145558_migrate_was_employed_in_claims.rb
```diff
@@ -0,0 +1,15 @@
+class MigrateWasEmployedInClaims < ActiveRecord::Migration[6.1]
+  class Claims < ActiveRecord::Base
+    self.table_name = :claims
+  end
+
+  def up
+    Claim.all.each do |was_employed|
```
garytaylor, line 7: There are a couple of issues here

1. You are looping over all claims and calling the item within the loop 'was_employed'.  This is very confusing when it is a claim ?
2. Once I replace the first 'was_employed' with 'claim' in my mind, you are doing 
```ruby

claim.update_columns was_employed: claim.was_employed.present?
```
This is definitely not right - go back to the original definition and work out what it should be

spec/forms/employment_form_spec.rb
```diff
@@ -98,13 +98,13 @@
...
 
-      it 'is false' do
-        expect(employment_form.was_employed).to be false
+      it 'is nil' do
+        expect(employment_form.was_employed).to be_nil
       end
     end
 
     context 'when the employment model has been persisted' do
-      before { allow(employment).to receive_messages persisted?: true }
```
garytaylor, line 107: Maybe change the name of this context as it no longer has anything to do with if the employment model is persisted ?

db/migrate/20220114143800_migrate_has_representative_in_claims.rb
```diff
@@ -0,0 +1,15 @@
+class MigrateHasRepresentativeInClaims < ActiveRecord::Migration[6.1]
+  class Claims < ActiveRecord::Base
+    self.table_name = :claims
+  end
+
+  def up
+    Claim.all.each do |representative|
```
garytaylor, line 7: Similar problem to another data migration here

1. Use of 'representative' as the variable name in the loop - but you are looping over claims when you look ?
2. You are effectively doing
```ruby

claim.update_columns has_representative: claim.has_representative.present?

```

Ill walk you through why this is wrong and why it is so important to check this value against what the calculated value was prior to this change.  If you were to run this, every single claim would have its has_representative field set to false.

1. You have added 'has_representative' to the claim schema - with no defaults, so the default is nil - every single claim at this point will have its has_representative set to nil
2. For each claim, your data migration sets the has_representative column to 'claim.has_representative.present?'.  As mentioned above, claim.has_representative will be nil at this point and 'nil.present?' will always return false.

app/views/claims/_employment.html.slim
```diff
@@ -1,7 +1,7 @@
 = f.fieldset label: t('.situation_legend')
   hr class="govuk-section-break.govuk-section-break--visible govuk-!-margin-bottom-3"
   /! Radios: main yes / no
-  = f.collection_radio_buttons :was_employed, optional: true, inline: true, include_hidden: true
+  = f.collection_radio_buttons :was_employed, optional: false, inline: true, include_hidden: true
```
garytaylor, line 4: You can remove optional: false as it defaults to false anyway

app/views/claims/_respondent.slim
```diff
@@ -11,7 +11,7 @@
 
 = f.fieldset label: { text: t('.workaddress_legend'), size: 'm' }
   hr class="govuk-section-break govuk-section-break--visible govuk-!-margin-bottom-3 govuk-!-margin-top-0"
-  = f.collection_radio_buttons :worked_at_same_address, optional: true, inline: true, include_hidden: false
+  = f.collection_radio_buttons :worked_at_same_address, optional: false, inline: true, include_hidden: false
```
garytaylor, line 14: You can remove optional: false as it defaults to false anyway

db/migrate/20220113132414_migrate_has_miscellaneous_information_in_claims.rb
```diff
@@ -0,0 +1,15 @@
+class MigrateHasMiscellaneousInformationInClaims < ActiveRecord::Migration[6.1]
+  class Claims < ActiveRecord::Base
+    self.table_name = :claims
+  end
+
+  def up
+    Claim.all.each do |claim|
+      claim.update_columns has_miscellaneous_information: claim.has_miscellaneous_information.present?
```
garytaylor, line 8: This does not look right - please check with the original calculated value.  When you look at it, you are setting 'has_miscellaneous_information' based on 'has_miscellaneous_information' - which will  be nil as it is a new column !

spec/forms/claim_details_form_spec.rb
```diff
@@ -46,13 +46,13 @@
 
   describe '#other_known_claimants' do
     context 'when #other_known_claimant_names is blank' do
```
garytaylor, line 48: this needs the context name changing - as it is no longer based on other_known_claimant_names

garytaylor (changes requested):
Hey Fred - there are a few issues that you need to resolve before this can be merged.  Please treat these data migrations as if your life depended on getting them right - it is very serious if we migrate data to incorrect values

garytaylor (approved):

PR #1484, opened 2022-05-18
RST-3818 - Fixed issue with form collection proxy when removing items

garytaylor:
**Before creating a pull request make sure that:**

- [ ] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [ ] tests have been updated / new tests has been added (if needed)

Please remove this line and everything above and fill the following sections:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/app/javascript/packs/components/RemoveMultiple.js
+++ b/app/javascript/packs/components/RemoveMultiple.js
@@ -15,6 +15,8 @@ function setupClickHandler() {
 
     e.preventDefault();
     removeMultiple(e.target.attributes['data-multiple-remove'].value)
+    showButton()
+    removePTag()
   });
 }
 
@@ -39,6 +41,16 @@ function removeMultiple(id) {
   multiple.style.display = 'none';
   markForDeletion(multiple);
   updateCount();
+}
+
+function showButton() {
+  const button = document.querySelector('button[data-multiple-add]')
+  button.classList.remove('hidden')
+}
+
+function removePTag() {
+  const p = document.getElementById('multiple-respondent-p')
+  p.remove()
 }
 
 function markForDeletion(multiple) {
--- a/app/javascript/packs/components/RemoveMultiple.js
+++ b/app/javascript/packs/components/RemoveMultiple.js
@@ -16,7 +16,7 @@ function setupClickHandler() {
     e.preventDefault();
     removeMultiple(e.target.attributes['data-multiple-remove'].value)
     showButton()
-    removePTag()
+    removeMultipleRespondentPTag()
   });
 }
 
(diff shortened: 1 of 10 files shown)


Fred83200 (approved):

app/javascript/packs/components/RemoveMultiple.js
```diff
@@ -15,6 +15,8 @@ function setupClickHandler() {
 
     e.preventDefault();
     removeMultiple(e.target.attributes['data-multiple-remove'].value)
+    showButton()
+    removePTag()
```
garytaylor, line 19: This needs a better name !  'Tag' means nothing :-)

app/javascript/packs/components/RemoveMultiple.js
```diff
@@ -41,6 +43,16 @@ function removeMultiple(id) {
   updateCount();
 }
 
+function showButton() {
+  const button = document.querySelector('button[data-multiple-add]')
+  button.classList.remove('hidden')
+}
+
+function removePTag() {
+  const p = document.getElementById('multiple-respondent-p')
```
garytaylor, line 52: There is a much easier way - mark anything in the html with data: { 'multiple-add': true } that needs hiding and showing.  Add the hidden class in the html where needed

Then, change the showButton function and name it showHiddenElements - then change the css selector to be '*[data-multiple-add]'. - you then won't need the removePTag function

app/javascript/packs/components/RemoveMultiple.js
garytaylor: but see my comment below first - you might not need this function

app/javascript/packs/components/RemoveMultiple.js
garytaylor: You will also need to change the document.querySelectorAll and loop over its results and remove the hidden class from all elements found

app/javascript/packs/components/RemoveMultiple.js
```diff
@@ -15,6 +15,8 @@ function setupClickHandler() {
 
     e.preventDefault();
     removeMultiple(e.target.attributes['data-multiple-remove'].value)
+    showButton()
+    removeMultipleRespondentPTag()
```
garytaylor, line 19: I know it doesn't seem important - but take another look at the name of this function.  Imagine you are someone new coming to the code base and you are looking at the page in your browser.  Does the name 'removeMultipleRespondetPTag' mean anything to you ?  it doesn't to me.

My guess is that you want to name it based on what the content you are removing actually is ?

garytaylor: Also, RemoveMultiple.js is not specific to multiple respondents - I believe it is used for both respondents and claimants - so unsure this is even the right place for it ?

PR #1488, opened 2022-07-28
Rst 4585

kalvinSolirius:
https://tools.hmcts.net/jira/browse/RST-4585

### Change description ###
Added error message for other type of claim in en and welsh.

Added validation for other type of claim.

Migration to make "is_other_type_of_claim" an attribute so value can be checked.

Ensured spec tests pass.

```
[ ] Yes
[ x ] No
```


db/migrate/20220729095120_update_is_other_type_of_claim_in_table_claims.rb
```diff
@@ -0,0 +1,8 @@
+class UpdateIsOtherTypeOfClaimInTableClaims < ActiveRecord::Migration[7.0]
+  def up
+    execute("UPDATE claims SET is_other_type_of_claim = true WHERE other_claim_details IS NOT NULL;")
```
garytaylor, line 3: This will only set them to true where other claim details is not null - the remaining claims will not have it set at all (i.e. the boolean column will still be set to NULL) - but this needs setting to false

garytaylor (changes requested):
Please see change regarding the migration

garytaylor (approved):

PR #1493, opened 2022-08-05
RST-4531

kalvinSolirius:

https://tools.hmcts.net/jira/browse/RST-4531

### Change description ###

Added code to clear text if radio button is false.

```
[ ] Yes
[ x ] No
```


garytaylor (changes requested):
Please can you add a spec for the form to prove the change that you have made - making sure that it resets it when it should and does not reset it when it shouldn't ?

spec/forms/claim_details_form_spec.rb
```diff
@@ -60,6 +61,17 @@
     end
   end
 
+  describe '#other_known_claimant_names' do
+    context 'when #other_known_claimant is false' do
+      it 'should not have a value' do
```
garytaylor, line 66: just a wording thing - maybe add 'after validation' to this ?  it is not clear that validation is involved

PR #1494, opened 2022-08-05
RST-4587

kalvinSolirius:

https://tools.hmcts.net/jira/browse/RST-4587

### Change description ###

Skip require_no_authentication when new or created to stop redirect to root.

Shouldn't, but needs further testing on isolated branch.
```
[ ] Yes
[ x ] No
```


garytaylor (changes requested):
Can you have a go at writing a feature / system spec that tests what this change actually does - i.e. going back in the browser history and making sure that the page number changes correctly etc.... ?

If you need any help with this please shout - I am not sure how much experience of capybara / rspec you have

spec/features/save_and_return_spec.rb
```diff
@@ -105,6 +105,25 @@
     end
   end
 
+  scenario 'returning to the existing application page from the claims details page' do
+    start_claim
+    saving_your_claim_page.register(password: 'green')
+    page.go_back
```
garytaylor, line 111: If you take a look at what the register method does, it looks like this

```ruby
def register(email_address: '', password:)
        email_address_question.set(email_address)
        memorable_word_question.set(password)
        save_and_continue_element.click
      end
```

But, what you might not know is that capybara does exactly as you say - so save_and_continue_element.click will click the save and continue element. It knows nothing about what should happen next so it cannot wait for anything to happen.
Therefore, it is important that you write your test as if YOU were testing it manually.  Imagine you are super fast like capybara - you wouldn't click the save and continue button THEN click back immediately afterwards - you would ensure that the new page Is shown before you click the back button.


Working in this way stops tests becoming sporadic.  For example, if the browser is particularly slow, when you click the back button it might not have even left the first page yet.  However, if the browser is very fast, the test may work as planned.

If you apply this strategy to all automated tests you write in any language, you will have nice solid tests :-)

garytaylor (changes requested):

garytaylor (approved):

PR #1497, opened 2022-08-10
Rst 4590

kalvinSolirius:

https://tools.hmcts.net/jira/browse/RST-4590

### Change description ###

Fixed additional claimants dob issue

```
[ ] Yes
[ x ] No
```


app/forms/additional_claimants_form.rb
```diff
@@ -36,8 +36,8 @@ class ClaimantForm < Form
     attribute :has_representative, :boolean
 
 
-    validates :date_of_birth, date: true, presence: true, comparison: {less_than: Date.today}
-
+    validates :date_of_birth, date: true, presence: true
+    validate :date_is_past
```
garytaylor, line 40: I would think that having the date validated to be in the past would be a really useful thing to use in other places in the app.  

Maybe we could move this to the date_validator ?

So, instead of

```ruby
validates :date_of_birth, date: true, presence: true
```

have

```ruby
validates :date_of_birth, date: { past: true }, presence: true
```

This syntax where we don't just have 'true' for the validator name ('date' in this case means date_validator) is passed to the initialize method of the date validator.  If you take a look at the date_validator.rb, you will see an 'omit_day' argument which is a good example of this pattern.

Another thing to look at is to use the date_range_validator - however, there is something important to consider when you want to pass in Date.today or anything which is not a static value (really, you should NOT use Date.today though - have a look at this if you want to know why - https://www.rubydoc.info/gems/rubocop/0.41.2/RuboCop/Cop/Rails/Date) - the problem is that when you call 'validates' like this in a class, it is executed at class definition time - so Date.today or Time.current or Time.zone.today etc... will ALL return the date of when the class was first created - i.e. when the application was deployed.
The answer is to use a lambda for the value(s).  The validator can then call the lambda when it comes to do the validation at run time.  The lambda would contain the code required to calculate the date - in this case Time.zone.today or whatever.

Lets look at what this could look like using the date_range_validator

```ruby

validates :date_of_birth, date: true, presence: true, date_range: { range: -> { .. Time.zone.yesterday } }

```

First of all, note that we have passed in the 'range' as a lambda (the -> syntax) - Im not sure how much ruby you know - so please ignore me if you already know this but a lambda is just like a function that can be stored in a variable or argument etc....  it can be called later using it's 'call' method.

That weird syntax with the '..' is a ruby range but because it has nothing in front of it, it means "define a range which covers anything from the beginning of time up to the value on the right hand side".  So, when the validator calls the 'cover?' method (check out the docs on the ruby 'Range') - any date that is UP TO yesterday will return true and therefore be valid.

However, this won't work straight away because the date_range_validator does not allow for the value to be a lambda.  However, this is easy to change - we could make it so that if the range responded to the 'call' method then it must be a lambda therefore to get the actual value we call the 'call' method.  If the range does not respond to 'call' then we just use the value as is (so existing code will work that does not use the lambda syntax)

Im not saying to definitely change this, but it would be useful, a little bit neater, but more importantly if you haven't had experience of custom validators and using lambdas etc.. it will give you a challenge.

app/validators/date_validator.rb
```diff
@@ -50,6 +53,15 @@ def illegal_year?(value)
     false
   end
 
+  def future_date?(value)
```
garytaylor, line 56: This doesn't make sense having a validator that has a 'past' argument that expects the caller to specify -> { Time.zone.today } - that could be hard coded and used when 'past' is set to false.  
If the validator is actually validating against a user provided date / time then maybe it should be called 'min' or 'max' instead of 'past'.  Makes more sense to the caller

garytaylor (changes requested):
Just a minor one - looks good though :-)

app/forms/additional_claimants_form.rb
```diff
@@ -36,8 +36,8 @@ class ClaimantForm < Form
     attribute :has_representative, :boolean
 
 
-    validates :date_of_birth, date: true, presence: true, comparison: {less_than: Date.today}
-
+    validates :date_of_birth, date: { in_the_past: true }, presence: true
+    validate :date_is_past
```
garytaylor, line 40: I don't understand why we have the date validator checking that the date is in the past AND a method called date_is_past as well doing the same thing ?

garytaylor (approved):

PR #1500, opened 2022-09-07
Rst 4590

kalvinSolirius:

https://tools.hmcts.net/jira/browse/RST-4590

### Change description ###

Added code to check age is between 10-100

Implemented new Error message in eng/welsh

Changed spec tests DoBs to 11.years.ago to ensure they don't break in future

Added spec tests for new age range validation

```
[ ] Yes
[ x ] No
```


app/validators/date_validator.rb
```diff
@@ -63,6 +66,15 @@ def future_date?(value)
     end
   end
 
+  def in_the_range?(value)
+    return false if within_range.nil?
+    if value.is_a?(String)
+      !((Date.parse(value) > 100.years.ago) and (Date.parse(value)) < 10.years.ago)
```
garytaylor, line 72: The range should not be in the validator - it should be passed in to the validator.  

The only alternative is to rename the validator to be something like

DateTenToOneHundredYearsAndMinimumValidator which obviously you don't want to do - my point is, the validator is doing something that the name doesn't tell us it is doing ?

app/validators/date_validator.rb
```diff
@@ -1,9 +1,10 @@
 # @TODO This class can be much simpler once we are not worrying about values being hashes,
 # strings etc.. (once all forms are converted to NullDbForm)
 class DateValidator < ActiveModel::EachValidator
-  def initialize(omit_day: false, in_the_past: nil, **kwargs)
+  def initialize(omit_day: false, in_the_past: nil, within_range: nil, **kwargs)
```
garytaylor, line 4: This validator is now both a date validator and a date range validator, I wonder if we should have 2 validators instead ?
The 'in_the_past' is also a range - i.e. the range is from the beginning of time until now

We did talk about this at the beginning but chose to go down this route and just accept an 'in_the_past' option.  However, we are now doing ranges as well, it is just making the responsibility of the validator too much.

I think that 

A DateValidator should validate the date - i.e. it is present and correct
A DateRangeValidator should validate the date range - i.e. it is between x date and y date
A DateRelativeValidator that can validate dates relative to today - this is where your 'in_the_past' should live.  Whilst this could be done as part of the date range validator, having a separate one will mean you can customise the messages better for dates relative to today

app/validators/date_range_validator.rb
```diff
@@ -1,7 +1,17 @@
 class DateRangeValidator < ActiveModel::EachValidator
+  def initialize(**kwargs)
```
garytaylor, line 2: Please can you add tests for the new validators.  If you have never done a test for a validator before, give me a shout - I can show you an example that I have just done

app/validators/date_range_validator.rb
```diff
@@ -1,7 +1,17 @@
...
+    super(**kwargs)
+  end
+
   def validate_each(record, attribute, value)
-    unless range.cover?(value)
-      record.errors.add(attribute, :date_range, start_date: start_date_str, end_date: end_date_str)
+    unless value.nil?
```
garytaylor, line 7: This level of indenting doesn't read nicely.

You could simplify this using guard clauses to return early under certain conditions

app/validators/date_range_validator.rb
```diff
@@ -1,7 +1,17 @@
...
+      unless range.cover?(value)
+        if date_format.nil?
+          record.errors.add(attribute, :out_of_range)
```
garytaylor, line 10: We have 2 different error codes - do we use both of them ?

app/validators/date_relative_validator.rb
```diff
@@ -0,0 +1,27 @@
...
+      record.errors.add(attribute, :less_than)
+    end
+  end
+
+  private
+
+  attr_reader :in_the_past
+
+  def future_date?(value)
+    return false if in_the_past.nil?
```
garytaylor, line 18: We really need to start running rubocop again - it will highlight things like this.

All guard clauses should have an empty line before the normal code begins - just makes for easier reading

app/validators/date_relative_validator.rb
```diff
@@ -0,0 +1,27 @@
...
+    time = Time.zone.today
+    if value.is_a?(String)
+      time < (Date.parse(value))
+    elsif value.is_a?(Date) || value.is_a?(Time)
+      time < (value)
```
garytaylor, line 23: You don't need these parentheses

spec/support/refunds_steps/setup_steps.rb
```diff
@@ -7,7 +7,7 @@ def given_i_am_luke_skywalker
   self.test_user = OpenStruct.new title: "Mr",
                                   first_name: "Luke",
                                   last_name: "Skywalker",
```
garytaylor, line 9: Not sure you need this ?  This is in the 'refunds' part of the application which you are not touching.

app/validators/date_range_validator.rb
```diff
@@ -1,22 +1,30 @@
...
-    end
+    return if value.nil? || range.cover?(value)
+
+    record.errors.add(attribute, :date_range, start_date: start_date_str, end_date: end_date_str)
   end
 
   private
 
   def start_date_str
+    return nil if date_format.nil?
```
garytaylor, line 15: would we not want a default string to be returned ?  i.e. the date in a 'standard' format - having these values as nil might cause issues if you had a message in the translation file that needed them to be strings ?

garytaylor (approved):

PR #1509, opened 2022-10-12
Rst 4650

kalvinSolirius:

https://tools.hmcts.net/jira/browse/RST-4650

### Change description ###

Added english and welsh redirects

```
[ ] Yes
[ x ] No
```


config/routes.rb
```diff
@@ -86,7 +86,8 @@
 
     get '/apply' => 'claims#new'
     get '/apply/refund' => 'refunds#new'
-    get '/apply/users', to: redirect('/apply/application-number')
```
garytaylor, line 89: I would have thought that you can just as easily get /apply/users - why not leave this in ?

garytaylor (approved):

PR #1512, opened 2022-11-23
RST-4803

kalvinSolirius:

https://tools.hmcts.net/jira/browse/RST-4803

### Change description ###

Changed other known claimant names presence to other known claimants

```
[ ] Yes
[ x ] No
```


db/migrate/20221123162401_change_other_known_claimants_to_be_radio_input.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def up
+    Claim.all.each do |claim|
+      claim.update_columns other_known_claimants: claim.other_known_claimants
+    end
+  end
+
+  def down
+    Claim.all.each do |claim|
```
garytaylor, line 13: I can't see a migration that adds the other_known_claimants column ?  Should that not be in this issue or am I missing something.  

This down block seems wrong anyway - but, it is probably not possible to do a down without knowing what the previous value was - unless you know it was nil of course ?

spec/views/claim_reviews/show/review_page_claim_details_spec.rb
```diff
@@ -27,7 +27,7 @@ def claim_path_for(*)
     end
 
     let(:claim) do
-      create :claim, claim_details: "wut\r\nwut", other_known_claimant_names: "Johnny Wishbone\r\nSamuel Pepys"
+      create :claim, claim_details: "wut\r\nwut", other_known_claimants: true, other_known_claimant_names: "Johnny Wishbone\r\nSamuel Pepys"
```
garytaylor, line 30: Not one for this spec - but there was nowhere else to add a comment.  You need to prove that the new attribute makes it to the API - there should already be a spec for that - but you will need to add this to it

db/migrate/20221123162401_change_other_known_claimants_to_be_radio_input.rb
```diff
@@ -0,0 +1,17 @@
+class ChangeOtherKnownClaimantsToBeRadioInput < ActiveRecord::Migration[7.0]
+  class Claim < ActiveRecord::Base
+    self.table_name = :claims
+  end
+
+  def up
+    Claim.all.each do |claim|
+      claim.update_columns other_known_claimants: claim.other_known_claimants
```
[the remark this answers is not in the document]
garytaylor: Now removed - this was not needed as it already exists from months ago

PR #1524, opened 2023-03-13
RST-5174

kalvinSolirius:


app/validators/special_character_validator.rb
```diff
@@ -0,0 +1,15 @@
+class SpecialCharacterValidator < ActiveModel::EachValidator
+
+  def validate_each(record, attribute, value)
+    return if value.nil?
+
+    if attribute == :address_building
```
garytaylor, line 6: It's not good practice to make decisions based on which attribute is using it.  If a particular attribute needs something slightly different, it is either a different validator - OR - passing options into the validator such as

```ruby
validates :address_building, special_character: { option1: true, option2: false, any_option: 4 }
```

these options are passed in to the initialize method of the validator

app/validators/special_character_validator.rb
```diff
@@ -0,0 +1,25 @@
+class SpecialCharacterValidator < ActiveModel::EachValidator
+
+  def initialize(**kwargs)
```
garytaylor, line 3: This initialize method is just calling super ?  That should mean it can be removed

spec/validators/special_character_validator_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    attribute :first_name,                :string
+    attribute :last_name,                :string
+    attribute :address_building,         :string
+    attribute :address_street,           :string
+    attribute :address_locality,         :string
+    attribute :address_county,           :string
+
+    validates :first_name, :last_name, special_character: true
+    validates :address_street, :address_locality, :address_county, special_character: true
```
garytaylor, line 47: Why have you got so many fields setup ?

Surely you just need 2 - one with special_character: true and another with special_character: { comma: true } ?

spec/validators/special_character_validator_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    model.attributes = valid_attributes
+    model.valid?
+    expect(model.errors).to be_empty
+  end
+
+  it 'is not valid for an input with special characters' do
+    model.attributes = invalid_attributes
+    model.valid?
+    model.errors.details.each do |error|
+      expect(error).to include(a_hash_including(error: :contains_special_characters))
```
garytaylor, line 62: Can you check that the error is on the attribute(s) that you expect as well ?

Check out the ActiveModel::Errors api - you can easily do things like model.errors.where(:attribute_name) etc...

PR #1532, opened 2023-05-17
RST-5337 Added validation length 14 for external postcodes and change…

kalvinSolirius:


app/forms/claimant_form.rb
```diff
@@ -1,16 +1,18 @@
...
+  EMAIL_ADDRESS_LENGTH     = 100
+  NAME_LENGTH              = 100
+  EXTERNAL_POSTCODE_LENGTH = 14
 
   include AddressAttributes.but_skip_postcode_validation
 
   validates :address_post_code,
             post_code: true, length: { maximum: POSTCODE_LENGTH },
             unless: :international_address?
+  validates :address_post_code, length: { maximum: EXTERNAL_POSTCODE_LENGTH }, if: :international_address?
```
garytaylor, line 15: Why do we only validate the length if it is an international address ?  Surely the length is important anyway, so there is no need to make this conditional ?

i.e. remove length: { maximum: POSTCODE_LENGTH } from validates :address_post_code and then have the length validation defined unconditionally ?

garytaylor (approved):

PR #1539, opened 2023-06-21
RST-5479 Changed email validation to strict

kalvinSolirius:


app/forms/claimant_form.rb
```diff
@@ -45,9 +45,9 @@ class ClaimantForm < Form
   validates :address_country, inclusion: { in: COUNTRIES }
   validates :fax_number,    presence: { if: :contact_preference_fax? }
   validates :email_address, presence: { if: :contact_preference_email? },
-                            email: { if: :contact_preference_email? },
                             ccd_email: { if: :contact_preference_email? },
                             length: { maximum: EMAIL_ADDRESS_LENGTH }
+  validates :email_address, email: { mode: :strict }, if: :contact_preference_email?
```
garytaylor, line 50: Why did this need its own line rather than adding mode: :strict into the existing email: {} block ?

spec/forms/claimant_form_spec.rb
```diff
@@ -151,6 +151,14 @@
       expect(claimant_form.errors).not_to include :mobile_number
     end
 
+    it 'disallows an invalid email' do
+      claimant_form.contact_preference = 'email'
+      claimant_form.email_address = 'jls435829@gmail..com'
```
garytaylor, line 156: Im not sure if this is the email address that I gave you as an example ?  If it is, please change it to something else as that was a real user.  If not, ignore me :-)

garytaylor (approved):

PR #1540, opened 2023-07-13
RST-5563

kalvinSolirius:
Added error translations


config/locales/en/et1/active_model.en.yml
```diff
@@ -219,12 +219,16 @@ en:
               blank: Say whether you’d like correspondence to be sent to the representative by post or email
             address_building:
               blank: Enter the building name or number from your representative’s address
+              contains_special_characters: Must not contain special characters
             address_street:
               blank: Enter the street from your representative’s address
+              contains_special_characters: Must not contain special characters
```
garytaylor, line 225: Rather than having multiple values exactly the same, you can refer to another value - so you can define it once and then refer to it everywhere else.

Here is an example

invalid_ccd_email: :'activerecord.errors.models.user.attributes.email_address.invalid'

garytaylor (approved):

PR #1542, opened 2023-07-27
RST-5558 changed href from anchor to js to fix redirect bug

kalvinSolirius:


app/views/guides/_writing_your_claim_statement.html.slim
```diff
@@ -48,4 +48,4 @@ article#writing_your_claim_statement
     li = t '.list_five.two'
 
   p.govuk-body
-    a.govuk-link href="#top" = t '.link_one'
+    a.govuk-link href="javascript:window.scrollTo({ top: 0 });" = t '.link_one'
```
garytaylor, line 51: In general, inline javascript is ugly and we don't use it.  
What you should do is follow the pattern of other pages where the page has its own javascript module which you call from the slim file the end.  The special link would need some attribute to identify it

Also, I am confused because the ticket says the 'Back to form' link - but this is the 'back to top' link ?

app/javascript/packs/application.js
```diff
@@ -35,6 +35,7 @@ import SessionPrompt from "./components/SessionPrompt";
 import RefundApplicantPage from "./pages/refunds/ApplicantPage";
 import RefundFeesPage from "./pages/refunds/FeesPage";
 import ClaimDetailsPage from "./pages/claims/ClaimDetailsPage";
+import WritingYourClaimStatementPage from "./pages/claims/WritingYourClaimStatementPage";
```
garytaylor, line 38: I thought we agreed that there was no such page as 'WritingYourClaimStatement' ?  If you look at the other javascript page modules, they all have the same name as the page where they are used.

We could invent like a 'component' module which we could use for partials, but it would need more thinking about as there may be more than one on a page, so ideally stick with page modules.

I can't remember what page this was used on else I would just say what to call it

garytaylor (approved):

PR #1544, opened 2023-08-18
RST-5294 Added sentry Context

kalvinSolirius:


app/controllers/application_controller.rb
```diff
@@ -60,6 +61,14 @@ def expired_session?
     false
   end
 
+  def set_sentry_context
+    return unless claim&.id.present? && claim&.application_reference.present? && claim.user&.id.present?
```
garytaylor, line 65: You have claim& in 2 out of the 3 conditions - but surely if you are expecting claim to possible be nil, it can be nil for all 3 ?

garytaylor (approved):

PR #1547, opened 2023-09-07
RST-5645 Added check for notice period end date is after start date

kalvinSolirius:


app/forms/employment_form.rb
```diff
@@ -37,6 +37,7 @@ class EmploymentForm < Form
   validates :new_job_start_date, date: true
   validates :notice_period_end_date, date: true
   validate :end_date_before_start_date?
+  validate :notice_period_end_date_before_start_date?
```
garytaylor, line 40: The name of this reads the wrong way around - it reads as "Validate that the notice period end date is before the start date"

spec/forms/employment_form_spec.rb
```diff
@@ -82,6 +82,17 @@
...
     end
+
+    context 'notice_period_end_date before start date' do
+
+      it 'rejects when notice_period_end_date is before start_date' do
+        employment_form.start_date = '1/1/2015'
+        employment_form.notice_period_end_date = '1/1/2014'
+
+        employment_form.valid?
+        expect(employment_form.errors.details[:notice_period_end_date]).to include(error: :notice_period_end_date_before_start_date)
```
garytaylor, line 93: Accessing errors as a hash like this is deprecated - you can use .errors.where - see the API documentation for ActiveModel::Errors for details

PR #1548, opened 2023-09-07
RST-5641 Added special character validation for work address

kalvinSolirius:


app/validators/special_character_validator.rb
```diff
@@ -1,17 +1,17 @@
...
-    if allow_comma == true && allow_numbers == true
-      regex = /^[a-zA-Z0-9 ,'-]+$/i
-    elsif allow_comma == true
-      regex = /^[a-zA-Z ,'-]+$/i
-    elsif allow_numbers == true
-      regex = /^[a-zA-Z0-9 '-]+$/i
-    else
-      regex = /^[a-zA-Z '-]+$/i
-    end
+    regex = if allow_comma == true && allow_numbers == true
```
garytaylor, line 6: This is not very scalable - having an if condition for each combination of options.

You could start off with a regex and add to it based on the conditions ?

PR #1549, opened 2023-09-08
(RST-5651) Initial fairly empty config for jenkins and helm chart

garytaylor:
### Change description ###

This is not the finished article - just an initial merge to get the jenkins file into master so that jenkins can see it

```
[ ] Yes
[X] No
```

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -1,0 +1,16 @@
+#!groovy
+
+@Library("Infrastructure")
+
+def type = "nodejs"
+def product = "et-pet"
+def component = "et1"
+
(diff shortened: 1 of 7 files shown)


Jenkinsfile_CNP
```diff
@@ -0,0 +1,16 @@
+#!groovy
+
+@Library("Infrastructure")
+
+def type = "nodejs"
```
cakeben, line 5: nodejs?

charts/et-pet-et1/values.aat.template.yaml
```diff
@@ -0,0 +1,4 @@
+nodejs:
```
cakeben, line 1: won't be nodejs either

PR #1593, opened 2023-10-31
RST-5641 Add character validation for work address

kalvinSolirius:


app/validators/special_character_validator.rb
```diff
@@ -1,17 +1,13 @@
...
-    if allow_comma == true && allow_numbers == true
-      regex = /^[a-zA-Z0-9 ,'-]+$/i
-    elsif allow_comma == true
-      regex = /^[a-zA-Z ,'-]+$/i
-    elsif allow_numbers == true
-      regex = /^[a-zA-Z0-9 '-]+$/i
-    else
-      regex = /^[a-zA-Z '-]+$/i
-    end
+    allowed_values = "a-zA-Z -'"
```
garytaylor, line 6: strings are frozen in later versions of ruby - and when we have #frozen_string_literals=true which we should have in our ruby files

Therefore the correct way of doing this is to use an array of strings - then use the .join method at the end to join the strings together

app/validators/special_character_validator.rb
```diff
@@ -1,17 +1,13 @@
 class SpecialCharacterValidator < ActiveModel::EachValidator
 
   def validate_each(record, attribute, value)
-    return if value.nil?
+    return if value.nil? || value.blank?
```
garytaylor, line 4: Isn't the responsibility of checking blank down to the 'allow_blank: true' option in the validates call ?  It doesnt seem right writing a validator that allows blank without you telling it to ?

app/validators/special_character_validator.rb
```diff
@@ -1,17 +1,13 @@
...
-    elsif allow_comma == true
-      regex = /^[a-zA-Z ,'-]+$/i
-    elsif allow_numbers == true
-      regex = /^[a-zA-Z0-9 '-]+$/i
-    else
-      regex = /^[a-zA-Z '-]+$/i
-    end
+    allowed_values = ["a-zA-Z -'"]
+    allowed_values << "0-9" if allow_numbers == true
+    allowed_values << "," if allow_comma == true
```
garytaylor, line 8: much nicer :-)

garytaylor (approved):

PR #1601, opened 2023-11-24
RST 6014 - fix: correctly displaying error messages when radio buttons are selected

lambley:


app/forms/respondent_form.rb
```diff
@@ -95,4 +93,8 @@ def reset_work_address!
   def reload_addresses
     target.addresses.reload
   end
+
+  def validate_acas_reason_presence
```
garytaylor, line 97: is this used ?

spec/forms/respondent_form_spec.rb
```diff
@@ -150,7 +155,7 @@
       describe 'when and ACAS number is given' do
         before { respondent_form.has_acas_number = 'true' }
 
-        it     { expect(respondent_form).not_to validate_presence_of(:no_acas_number_reason) }
```
garytaylor, line 153: why is it not important to prove that we are not validating no_acas_number_reason in this scenario ?

app/forms/respondent_form.rb
lambley: no, should have removed that. removed now.

spec/forms/respondent_form_spec.rb
lambley: would this be a better alternative? The test cases pass with this change:

```
describe 'when an ACAS number is given' do
  before { respondent_form.has_acas_number = 'true' }

  it     { expect(respondent_form).to validate_presence_of(:acas_early_conciliation_certificate_number) }
  it     { expect(respondent_form).to_not validate_presence_of(:no_acas_number_reason) }
end

describe 'when an ACAS number is given' do
  before { respondent_form.has_acas_number = 'false' }

  it     { expect(respondent_form).to validate_presence_of(:no_acas_number_reason) }
  it     { expect(respondent_form).to_not validate_presence_of(:acas_early_conciliation_certificate_number) }
end
```

garytaylor: yes that looks right

PR #1644, opened 2024-09-12
RST-6870 Validation for claimant other title presence

kalvinSolirius:


app/forms/claimant_form.rb
```diff
@@ -76,6 +77,10 @@ def last_name=(name)
 
   private
 
+  def other_title_selected?
+    title == I18n.t("claims.personal_details.title.options.Other")
```
garytaylor, line 81: Have you tested this in welsh ?  I don't think it will work.

The dropdown should be configured so that the key from the translation file is what is stored in the database, not the value shown to the user which is the value from the translation file

If you look, the keys are the same irrespective of language - but your code will translate to 'Arall' which will return false

spec/forms/claimant_form_spec.rb
```diff
@@ -68,6 +68,20 @@
       end
     end
 
+    describe 'other_title' do
+      before { claimant_form.title = 'Other' }
```
garytaylor, line 72: You are setting the title outside of the context named 'when title is other' which doesn't read so well.

garytaylor (approved):

